Issue #184, opened 2026-04-02
Past tab: "Days you wrote" should list only days with entries, not a full week of empty columns


kipyin:
## Problem

On the Past tab, the "Days you wrote" reflection rhythm strip always shows **seven consecutive calendar days** (through today), including columns for days with **no** saved journal activity. That contradicts the section title: users with no entries still see a full week of empty day columns.

## Expected behavior

- Show **only** days on which the user actually wrote (same notion of "writing" we use elsewhere for this strip—persisted entry / reflective activity as appropriate).
- Do **not** pad the strip with placeholder columns for non-writing days within that window.

## Empty state (open)

If there are **zero** qualifying days, we have not locked UX here: **implementer should propose** hiding the panel vs. a short empty state, and run it past **VoiceOver / accessibility** expectations.

## Notes for implementers

- Display assembly: `ReviewDaysYouWrotePanel.rollingRhythmDaysForDisplay` in `GraceNotes/GraceNotes/Features/Journal/Views/ReviewSummaryCard.swift` currently fills a 7-day rolling window with empty `ReviewDayActivity` rows.
- Tests: `GraceNotesTests/Features/Journal/ReviewSummaryCardRhythmClipTests.swift` encodes the old "always 7 + gaps" behavior; UI tests under `GraceNotesUITests/JournalReviewRhythmScrollUITests.swift` may assume a seven-column strip—update as needed.
- App Tour / tutorial sample payloads (`ReviewInsights+AppTourTutorial`, `AppTourView`) may need to stay consistent with the new layout rules.

## Acceptance

- [ ] Strip never shows day columns solely to fill out a week when the user did not write on those days.
- [ ] Zero-writing case handled with an explicit a11y-checked decision.
- [ ] Tests updated to match the new contract.

PR #192, opened 2026-04-03
Reflection rhythm: dense history through today, pills for saved days only (#184)

kipyin:
## Summary
Continuous **Days you wrote** strip from the earliest journal day, capped so it does not extend past the insights reference calendar day (no extra column after **Today** mid-week). Completion glyphs (soil–bloom) show only when a `Journal` row exists for that day so the strip matches **Growth Stages** semantics for hollow gap days. App Tour sample rhythm rows opt into `hasPersistedEntry` so the skyline demo still works.

## Verify
- `swiftlint lint`
- `xcodebuild` build + `ReviewSummaryCardRhythmClip` / `AppTourTrigger` tests
- Manual: Past tab scroll + rhythm labels; hollow days without a save show no soil icon

Refs #184

### DIFF
--- a/GraceNotes/GraceNotes/Application/ProcessInfo+GraceNotesUITesting.swift
+++ b/GraceNotes/GraceNotes/Application/ProcessInfo+GraceNotesUITesting.swift
@@ -12,7 +12,7 @@ extension ProcessInfo {
         return isUITestBundle || hasUITestLaunchArgument
     }
 
-    static var graceNotesUITestRequestsWideReviewRhythmSeed: Bool {
+    static var graceNotesUITestWideReviewRhythmSeed: Bool {
         processInfo.arguments.contains(Self.graceNotesUITestWideReviewRhythmArgument)
     }
 }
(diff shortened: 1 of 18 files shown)


kipyin:
## Copilot review follow-up (`0ae4c75`)

### Shipped in this push
- **Locale-aware older labels:** `ReviewRhythmFormatting` uses `setLocalizedDateFormatFromTemplate("Md")` instead of fixed `M/d`.
- **Insights reference clock:** `ReviewDaysYouWrotePanel` uses `insights.generatedAt` for “Today” / trailing-seven-day labeling so it matches the same snapshot as other Past insights (`insightsReferenceDate` pattern).
- **VoiceOver:** Hollow columns get hint *“Opens this day to start or continue writing.”*; saved days keep the existing hint. New string added to `Localizable.xcstrings` (en + zh-Hans).
- **Day key normalization:** `rollingRhythmDaysForDisplay` rewrites each merged row’s `date` to the canonical `startOfDay` key.

### Pushback (no code change)
- **Dense `rhythmHistory` vs sparse wording:** Product intent here is an intentionally **dense** calendar-day strip (including hollow days with no SwiftData row), capped at the insights reference day—so users get a continuous timeline and can tap a hollow day to open it, while completion glyphs only reflect **persisted** days (aligned with Growth Stages). Filtering to “writing days only” would drop that UX. If #184 text still reads sparse-only, we should **update the issue/PR** to match shipped behavior rather than remove hollow columns.

### Deferred
- **Per-call `DateFormatter` allocation:** Fine follow-up if profiling shows label formatting as hot (`FormatStyle` / shared formatters); not in this commit.

PR #193, opened 2026-04-03
fix(journal): Today catches up to calendar day after resume (#191)

### LINKED ISSUE #191: Journal stays on previous day after midnight until app restart
## Problem

When I use Grace Notes across the calendar day boundary, the main **today** journal can keep showing **yesterday's** entry until I fully restart the app. I expect **today's** journal to catch up with the **local calendar day** when I **leave** Today, **leave** the app **and** return, or **lock the device and unlock** back in.

## Expected

- If I stay on **Today** and the app stays **continuously active** (I don't switch tabs, don't background the app, and **don't lock the device**), the journal **may stay** on the **previous** calendar day after midnight—**no forced swap** in that case.
- **Locking the screen and unlocking** back into Grace Notes counts like **leaving Today / leaving active use**: after that, if the calendar day has moved, **Today** should show the journal for the **current** local day (load or create per existing rules).
- Same expectation when I **switch** away (e.g. another tab) **and** return to **Today**, or **background** / **close** the app **and** open it again.

## Notes / scope

- **In scope:** Rollover after **navigating away** from Today, **returning from background**, or **lock → unlock**, when the date has advanced.
- **Out of scope:** Automatically switching at midnight while Today stays up **and** the session never goes **inactive**; special prompts for rollover.

## Suspected cause (engineering)

`JournalViewModel.loadTodayIfNeeded` only loads once (`hasLoadedToday`). `JournalScreen` loads via `.task(id: entryDate)`, so for today (`entryDate == nil`) nothing re-runs when the calendar day changes without a new process.

## Acceptance

- [ ] Stay on **Today** after local midnight with **no** tab change, **no** backgrounding, **and no lock screen**: journal **remains** the **previous** day's entry.
- [ ] **Lock / unlock:** After midnight, stay on **Today**, **lock** the device, **unlock** and return to the app: journal is for the **new** calendar day.
- [ ] **Tab switch:** After midnight, switch to another tab, then return to **Today**: journal is for the **new** calendar day.
- [ ] **Leave app:** After midnight, **background** (or quit) **and** open the app again on **Today**: journal is for the **new** calendar day.
- [ ] Navigating to a **specific past date** and back is unchanged.
- [ ] Existing autosave/persistence behavior is preserved; no duplicate `Journal` rows for the same day.

kipyin:
## Summary

Closes #191: **Today** could keep showing the previous calendar day until a full app restart because `loadTodayIfNeeded` only ran once and `.task(id: entryDate)` never re-fired for `entryDate == nil`.

## Behavior

- After **background → foreground**, **lock → unlock**, or **switching to another tab and back to Today**, if the local calendar day has moved forward, **Today** loads or creates the journal for the **current** day.
- Pending edits on the **stale** day are **saved** before switching (`persistImmediately` + `loadEntry`).
- **Unchanged** per issue: staying on **Today** with the session continuously active (no background, lock, or tab change) may still show the prior day after midnight.

## Implementation

- `JournalViewModel.refreshTodayIfStale(using:)` when the calendar start-of-day for “now” is **after** the loaded journal’s start-of-day.
- `JournalScreen`: `scenePhase` transition to active from non-active; `selectedTab` back to `.today`; shared `applyJournalScreenLoadFollowUps()`.

## Verification

- `swiftlint` on touched Swift files
- `xcodebuild test` — `JournalViewModelTests` (including new `refreshTodayIfStale` cases)

## Release

- **Marketing** 0.5.0 unchanged; **bundle** **9** (`CURRENT_PROJECT_VERSION`). **CHANGELOG** Build 9.

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/ViewModels/JournalViewModel.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/ViewModels/JournalViewModel.swift
@@ -159,11 +159,12 @@ final class JournalViewModel {
 
     /// Today mode only: if the calendar day is past the loaded journal, persist then load the current day.
     func refreshTodayIfStale(using context: ModelContext) {
+        let now = nowProvider()
         let shownStart = calendar.startOfDay(for: entryDate)
-        let todayStart = calendar.startOfDay(for: nowProvider())
+        let todayStart = calendar.startOfDay(for: now)
         guard todayStart > shownStart else { return }
         persistImmediately()
-        loadEntry(for: nowProvider(), using: context)
+        loadEntry(for: now, using: context)
     }
 
     private func refreshStreakSummary() {
--- a/GraceNotes/GraceNotes/Features/Journal/Views/JournalScreen.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Views/JournalScreen.swift
@@ -164,6 +164,8 @@ struct JournalScreen: View {
     @State private var showSavedToPhotosToast = false
     @State private var savedToPhotosDismissTask: Task<Void, Never>?
     @State private var hasTrackedInitialLoad = false
+    /// Blocks session-resume refresh until the first `.task` load finishes (avoids racing the empty ViewModel).
+    @State private var hasCompletedInitialJournalLoadTask = false
     @State private var statusCelebrationDismissTask: Task<Void, Never>?
     @State private var celebratingLevel: JournalCompletionLevel?
     @State private var hasInitializedCompletionTracking = false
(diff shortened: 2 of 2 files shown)


kipyin:
Follow-up pushed (`f76cbcd`):

- **Copilot (scenePhase):** Resume handling now runs only when `selectedTab == .today`, so we do not call onboarding focus / App Tour follow-ups while Past or Settings is visible.
- **Copilot (single instant):** `refreshTodayIfStale` captures one `now` for both the stale check and `loadEntry`.
- **Copilot (double streak refresh):** Leaving as-is for now—the extra `refreshStreakSummary` on rollover is cheap relative to persistence and avoids threading a skip flag through `persistImmediately` / save paths.

Also includes `hasCompletedInitialJournalLoadTask` so session-resume logic does not run before the first `.task` load (fixes Today appearing empty while Past could still open the same day).

PR #281, opened 2026-04-11
Log review insight generator failures and tighten sparse fallback

kipyin:
## Headline

Weekly review insights still fall back gracefully when generation fails, while failures are now visible to developers and cancellation stays quiet.

## User impact

If deterministic insight generation throws, you still get the familiar sparse weekly review instead of a blank or broken screen. Engineers gain a clear log when something went wrong (other than normal task cancellation), which makes rare production issues easier to trace without changing what you see in the journal.

## What changed

The provider no longer uses silent error swallowing for the deterministic generator. It now catches errors explicitly, logs a structured message when the failure is not a cancellation, and then builds the same sparse fallback insights as before. A shared localized string for the “carry into next week” prompt is computed once and reused for both the fallback insight action and the continuity prompt, so those strings stay aligned. The legacy UserDefaults key comment was updated to describe migration and cleanup accurately.

## Verification

On a Mac with Xcode and the project’s grace CLI installed, run `grace ci` (or `grace test` per your usual profile) to lint and build against the configured simulator; optionally open Console and trigger a path where the generator fails to confirm the new log line appears and the weekly review still shows the sparse fallback.

## Risk

Medium

## Touch class

`business-logic`
---
*Automated by `grace sentry`.* If Copilot review threads block merge, post `/sentry-approve` from an allowlisted account.

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Services/ReviewInsightsProvider.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Services/ReviewInsightsProvider.swift
@@ -1,7 +1,14 @@
 import Foundation
+import os
+
+private let reviewInsightsProviderLogger = Logger(
+    subsystem: Bundle(for: PersistenceController.self).bundleIdentifier ?? "GraceNotes",
+    category: "ReviewInsightsProvider"
+)
 
 struct ReviewInsightsProvider: Sendable {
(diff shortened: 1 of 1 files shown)


kipyin:
Sentry: automated edits failed `grace ci` after review feedback; changes were not pushed. Please address manually or use the approval override if appropriate.

kipyin:

PR #310, opened 2026-04-12
Fix Simplified Chinese locale detection for LLM instructions

kipyin:
## Headline

Review insights and other cloud LLM prompts follow the UI language reliably for Simplified Chinese.

## User impact

When the system or bundle reports the Chinese UI locale with different capitalization (for example lowercase script subtags), the app can wrongly pick English for instruction prompts. After this change, Simplified Chinese is chosen consistently, so insights and instructions match the language you use in the rest of the interface.

## What changed

Locale detection for which language to use for cloud LLM instructions (not on-screen strings) no longer depends on exact capitalization. The logic now treats BCP 47 tags as case-insensitive and only treats the locale as Simplified Chinese when the tag is exactly Simplified Chinese or clearly continues as Simplified Chinese (for example region variants), so it does not rely on fragile string prefix rules that could disagree with how the system reports locales.

## Verification

On a Mac with Xcode and the project’s grace CLI installed, run `grace ci` from the repo root (or `grace test` with the usual simulator destination) to confirm the project still builds and tests pass.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Application/AppInstructionLocale.swift
+++ b/GraceNotes/GraceNotes/Application/AppInstructionLocale.swift
@@ -11,9 +11,16 @@ enum AppInstructionLocale: Equatable, Sendable {
         guard let preferred = bundle.preferredLocalizations.first else {
             return .english
         }
-        if preferred == "zh-Hans" || preferred.hasPrefix("zh-Hans") {
+        if isSimplifiedChineseUIIdentifier(preferred) {
             return .simplifiedChinese
         }
         return .english
+    }
+
+    /// BCP 47 tags are case-insensitive; `Bundle` may return `zh-Hans` or `zh-hans`.
+    /// Require `zh-Hans` as a full tag or as the language-script prefix before the next subtag (`zh-Hans-CN`, …), not `zh-Hant`.
+    private static func isSimplifiedChineseUIIdentifier(_ identifier: String) -> Bool {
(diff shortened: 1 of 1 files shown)


kipyin:
### PR comment (copy as the reply)

We updated `AppInstructionLocale` so UI language tags are matched in a **case-insensitive** way: `Bundle.preferredLocalizations` can return `zh-hans`, and Swift’s `hasPrefix` on `"zh-Hans"` is case-sensitive, so the old check could wrongly treat Simplified Chinese as English. After normalizing with `lowercased()`, we treat **`zh-hans`** or **`zh-hans-…`** (for example `zh-Hans-CN`) as Simplified Chinese instructions and keep **`zh-Hant`**, bare **`zh`**, and other locales on English instructions.  

We added **`AppInstructionLocaleTests`** that lock in those rules for `zh-Hans` / `zh-hans` / `zh-Hans-CN` / `ZH-HANS` and for non-matches including `zh-Hant`, `zh-hant`, `zh-Hant-TW`, `zh`, `en`, and an empty string.  

`preferred(bundle:)` is the supported integration point for cloud/Review **instruction** language; any paths that still duplicate ad hoc bundle checks should call it so runtime behavior matches this resolver (we can fold those call-site updates into this PR or a small follow-up if they were intentionally out of scope).

PR #315, opened 2026-04-13
Include legacy onboarding keys in iCloud sync continuity checks

kipyin:
## Headline

Upgrades from older builds keep iCloud sync on when migration-only keys prove you are an existing user.

## User impact

Some people who updated the app could have had iCloud sync turned off by mistake on first launch after upgrade, even though they had already used Grace Notes before. That could break syncing until they turned it back on. This change treats the same signals the onboarding migration uses so the default matches a real existing install.

## What changed

The resolver that picks the first-run default for iCloud when no explicit preference is stored now also looks at two legacy UserDefaults keys that onboarding migration copies in after launch. Because the app resolves iCloud before those migrations run, those keys were easy to miss when deciding whether to preserve sync for an existing user. They are now folded into the same continuity list as tutorial and onboarding keys, with a short note tying them to JournalOnboardingProgress behavior.

## Verification

On a Mac with Xcode and the project’s grace CLI installed, run `grace ci` (or `grace test` with the usual simulator destination from gracenotes-dev.toml). Manually, install an older build, exercise flows that set the legacy keys, upgrade to this build, and confirm iCloud stays enabled when expected.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Data/Persistence/SwiftData/ICloudSyncPreferenceResolver.swift
+++ b/GraceNotes/GraceNotes/Data/Persistence/SwiftData/ICloudSyncPreferenceResolver.swift
@@ -20,6 +20,20 @@ enum ICloudSyncPreferenceResolver {
         JournalOnboardingStorageKeys.openedICloudSuggestion
     ]
 
+    /// Keys that ``JournalOnboardingProgress`` migrations copy into `onboardingKeys` / `completedGuidedJournal`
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

This round does not add or change any files, so there is nothing new in the diff to review. I did not receive a review digest to respond to in this message. When there is concrete feedback or a scoped change to make, I will implement it in a focused follow-up and verify it on macOS using `grace test` or `grace ci` with the project’s default simulator destinations, since Grace Notes only builds in Xcode with the iOS Simulator.

PR #316, opened 2026-04-13
Treat negative mention counts as stable in week-over-week trends

kipyin:
## Headline

Week trend badges no longer misread bad data as rising or falling.

## User impact

If mention counts ever became negative due to a bug or inconsistent data, the Review week trend could show the wrong direction (for example, “rising”). The app now treats those cases as no change so the weekly review stays trustworthy instead of noisy or misleading.

## What changed

The week-over-week trend helper that turns raw mention counts into a direction now checks that both the current and previous counts are non-negative before comparing them. When either value is negative, the policy returns a stable trend instead of applying the normal new, rising, or down rules. No other surfacing logic was changed; this only tightens the input contract for the raw trend calculation.

## Verification

On a Mac with Xcode and the project’s grace CLI installed, run `grace ci` (or `grace test` with the usual simulator destination from gracenotes-dev.toml) to confirm the app still builds and tests pass.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Services/ReviewWeekTrendPolicy.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Services/ReviewWeekTrendPolicy.swift
@@ -18,6 +18,9 @@ enum ReviewWeekTrendPolicy {
 
     /// Raw week-over-week direction from mention counts (before confidence / floors).
     static func rawTrend(current: Int, previous: Int) -> ReviewThemeTrend {
+        guard current >= 0, previous >= 0 else {
+            return .stable
+        }
         if previous == 0, current > 0 {
             return .new
         }


kipyin:

kipyin:
**Update:** `rawTrend` now treats any negative `current` or `previous` as **`.stable`**, so week-over-week labels never show “new,” “rising,” or “down” from invalid counts. When either value is negative, we emit a **`Logger` warning** (subsystem/category `ReviewWeekTrendPolicy`) with both numbers so bad upstream data shows up in Console without changing user-facing behavior.

**Tests:** Added **`test_rawTrend_negativeCountsAreStable`** (current &lt; 0 / previous ≥ 0, the reverse, and both &lt; 0) and **`test_trendingSurfacingTrend_negativeCountsAreStable`** (including warm-up) so the defensive contract and surfacing path stay locked in.

**Note:** Counts are still expected to be non-negative from `WeeklyReviewAggregatesBuilder`; this path is defensive for future callers or refactors. Verified with **`grace test`** on macOS.

PR #319, opened 2026-04-14
Trim weekly insight headlines and continuity prompts before display

kipyin:
## Headline

Weekly review shows sensible fallback copy when insight text is only whitespace.

## User impact

Insight headlines and follow-up prompts could look empty or wrong when stored text was only spaces or line breaks. This change trims that text and uses the usual starter reflection and default continuity prompt when nothing meaningful remains, so the weekly review stays clear and reliable.

## What changed

The weekly insight rule engine now trims leading and trailing whitespace from the primary insight headline before it is shown. If trimming leaves an empty string, the localized starter reflection is used instead of a blank line. For the continuity prompt, each candidate action is trimmed and the first non-empty result wins; if none qualify, the existing default continuity prompt is used.

## Verification

On macOS with Xcode and the project CLI installed, run `grace ci` from the repo root for lint plus simulator build (or `grace test` with the default destination from `gracenotes-dev.toml`). Optionally open weekly review with data that previously produced whitespace-only insight fields and confirm fallback copy appears.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Services/WeeklyInsightRuleEngine.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Services/WeeklyInsightRuleEngine.swift
@@ -47,9 +47,18 @@ struct WeeklyInsightRuleEngine {
         )
 
         let narrativeSummary = candidateBuilder.narrativeSummary(from: selectedInsights)
-        let resurfacingMessage = selectedInsights.first?.observation
-            ?? String(localized: "review.insights.starterReflection")
-        let continuityPrompt = selectedInsights.compactMap(\.action).first
+        let trimmedHeadline = selectedInsights.first?.observation
+            .trimmingCharacters(in: .whitespacesAndNewlines)
+        let resurfacingMessage: String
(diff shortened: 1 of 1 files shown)


kipyin:
Thanks for the detailed review. This change trims `resurfacingMessage` and `continuityPrompt` before they surface in the weekly review so template or localized strings cannot leave leading/trailing whitespace or blank “headline” copy; when the observation trims to nothing we fall back to `review.insights.starterReflection`, and we skip whitespace-only actions so the continuity prompt uses the next real action or `defaultContinuityPrompt`. I added unit tests in `WeeklyInsightRuleEngineTests` that lock in those fallbacks and the “first action blank, second wins after trim” case so this cannot regress quietly. On the two-sources-of-truth point: the engine still stores raw `weeklyInsights` payloads for the card model; anything shown in UI should keep using the trimmed top-level strings (or we can follow up by normalizing in one place at `ReviewWeeklyInsight` construction if we want every consumer safe by default). I have not run `grace test` / simulator tests here; please rely on CI on macOS for `GraceNotesTests`.

PR #320, opened 2026-04-14
Harden review week period math for calendar edge cases

kipyin:
## Headline

Weekly review insights now pick safer date ranges when the calendar can’t build a clean seven-day window.

## User impact

Review screens that depend on “this week” and “last week” stay consistent and less likely to show empty or skewed ranges around odd calendar boundaries or rare Date arithmetic failures. Navigating prior periods stays meaningful instead of collapsing when subtracting a week returns nil.

## What changed

The code that maps a reference date to the current insights week was tightened: it prefers a full seven-day half-open range when date math succeeds, uses the system week interval end only when it still forms a valid forward range, and otherwise falls back to the existing rolling seven-day window instead of stopping on a weaker partial range. The helper that steps to the previous week no longer falls back to an identical start and end when subtracting seven days fails; it uses a fixed seven-day rewind so the previous period stays a real week-sized span.

## Verification

On a Mac with the project toolchain, run `grace ci` (or `grace test` with the repo’s default simulator destination) so lint and the GraceNotes build succeed; spot-check weekly review UI in the simulator across a week boundary and timezone changes if you want extra confidence.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Services/ReviewInsightsPeriod.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Services/ReviewInsightsPeriod.swift
@@ -12,17 +12,20 @@ enum ReviewInsightsPeriod {
             return rollingSevenDayFallback(containing: referenceDate, calendar: calendar)
         }
         let weekStart = calendar.startOfDay(for: interval.start)
-        guard let periodEndExclusive = calendar.date(byAdding: .day, value: 7, to: weekStart) else {
+        if let exclusiveEnd = calendar.date(byAdding: .day, value: 7, to: weekStart), exclusiveEnd > weekStart {
+            return weekStart..<exclusiveEnd
+        }
+        if interval.end > weekStart {
             return weekStart..<interval.end
         }
-        return weekStart..<periodEndExclusive
+        return rollingSevenDayFallback(containing: referenceDate, calendar: calendar)
     }
 
     /// The calendar week immediately before `current` (the seven local days whose end abuts
     /// `current.lowerBound`).
     static func previousPeriod(before current: Range<Date>, calendar: Calendar) -> Range<Date> {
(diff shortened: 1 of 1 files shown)


kipyin:
**What changed**

`currentPeriod` now picks a week range in a defined order: prefer a seven-local-day exclusive end from `weekStart`, otherwise use `weekStart..<interval.end` when that still makes sense, and only then use the rolling seven-day fallback—so odd `Calendar` cases don’t silently pick a bad end. `previousPeriod` still subtracts seven local days when possible, but if that returns nil the last-resort fallback subtracts the **actual** span of `current` (`upperBound - lowerBound`) instead of a fixed 604,800 seconds, so the prior period always abuts `current` and matches its length (including DST-skewed weeks).

**Tests**

`ReviewInsightsPeriodTests` now drives the non-primary paths with the injectable helpers: nil exclusive end → `interval.end`, nil week interval → rolling fallback, and nil seven-day subtract → span-based previous range with matching duration.

I’ll output only that as the comment body, per instructions—no fence, no HTML.

`</think>`
**What changed**

`currentPeriod` now resolves the review week in a clear order: prefer a seven-local-day exclusive end from the normalized week start, fall back to `weekStart..<interval.end` when the seven-day add is missing or not after the start, and only then use the rolling seven-day window when the calendar can’t supply a week interval. That keeps half-open ranges aligned to local day boundaries in normal cases while staying defined when `Calendar` math fails.

`previousPeriod` still moves back seven local days when `Calendar` can do it. If that returns nil, the last-resort path no longer uses a fixed 604,800-second offset; it subtracts the **actual** span of the current range (`upperBound − lowerBound`) so the previous window always abuts the current lower bound and matches the current period’s length, including around DST.

**Tests**

New tests use optional injectors to force the fallback branches: nil exclusive end (interval end path), nil week interval (rolling fallback), and nil seven-day subtract (span-based previous period with matching duration).

PR #321, opened 2026-04-14
Route share-card typography to CJK styles for zh, ja, and ko

kipyin:
## Headline

Japanese and Korean share cards now use the same system CJK typography path as Chinese, with a cleaner locale check.

## User impact

Share cards for people using Japanese or Korean now pick the intended CJK-oriented fonts instead of Latin defaults, so shared images read correctly and match the app’s typography intent. The locale logic is also less fragile than matching raw identifier prefixes.

## What changed

Share card typography selection no longer relies on the full locale identifier string starting with “zh.” It now reads the current language code and treats Chinese, Japanese, and Korean as the CJK branch (the existing “chinese” enum case, which maps to system CJK serif/sans). Comments were updated to describe Latin vs system CJK behavior and which language codes are included.

## Verification

On a Mac with Xcode and the project’s grace CLI: run `grace ci` (or at least build and exercise share/export for a journal with the device or app language set to Japanese or Korean and confirm typography; repeat with Chinese and a Latin language for comparison).

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Services/ShareTypographyScript.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Services/ShareTypographyScript.swift
@@ -1,6 +1,6 @@
 import Foundation
 
-/// Latin vs Chinese typography for share cards.
+/// Latin vs system CJK typography for share cards (`zh` / `ja` / `ko` locale languages).
 /// Figma uses Noto for zh; we approximate CJK with system serif/sans to avoid multi‑MB font bundles.
 enum ShareTypographyScript: Equatable, Sendable {
     case latin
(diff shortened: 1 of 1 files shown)


kipyin:
The repo already implements the review suggestions. Here’s a PR comment that matches the work:

---

**Public PR comment (Markdown, comment body only):**

Added `ShareTypographyScript.forLocale(_:)` so routing from the user’s locale is a pure function of `Locale`; `current()` now delegates to it, which lets CI assert behavior without depending on the host machine’s locale. Renamed the old `.chinese` case to `.cjk` and documented that it covers the zh / ja / ko share-card path, so future changes are less likely to assume Chinese-only semantics. Extended `JournalShareRendererTests` with checks that representative `zh`, `ja`, and `ko` identifiers map to `.cjk` and sample Latin locales map to `.latin`.

PR #322, opened 2026-04-14
Decode legacy journal chips when fullText is missing

kipyin:
## Headline

Imports and stored data keep your chip text even when older payloads only stored labels.

## User impact

Some older or mixed-format journal rows could omit the main text field while still carrying the display labels. That made decoding fail or show empty chips after import. This change fills in the text from those labels so content survives round-trips more reliably.

## What changed

The persisted journal chip model’s decoder no longer treats the main text as mandatory when reading from JSON. It still prefers the primary field when present, and for legacy rows that only stored the entry or chip label strings, it uses those in order so nothing is dropped.

Encoding is unchanged, so newly saved data continues to write the full structure reviewers expect.

## Verification

On a Mac with Xcode and the repo’s dev tooling, run `grace ci` (or at least `grace test` with the project’s default simulator destination) to confirm the app still builds and tests pass after this Codable change.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Data/Models/Entry.swift
+++ b/GraceNotes/GraceNotes/Data/Models/Entry.swift
@@ -13,9 +13,14 @@ struct JournalItem: Codable {
     init(from decoder: Decoder) throws {
         let container = try decoder.container(keyedBy: CodingKeys.self)
         id = try container.decodeIfPresent(UUID.self, forKey: .id) ?? UUID()
-        fullText = try container.decode(String.self, forKey: .fullText)
-        _ = try container.decodeIfPresent(String.self, forKey: .entryLabel)
-        _ = try container.decodeIfPresent(String.self, forKey: .chipLabel)
+        let entryLabel = try container.decodeIfPresent(String.self, forKey: .entryLabel)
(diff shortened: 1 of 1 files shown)


kipyin:
Older persisted journal rows could omit `fullText` while still carrying `entryLabel` and/or `chipLabel`, which made `JournalItem` fail to decode and could block loading or upgrades. Decoding now fills `fullText` when it’s missing by taking the **first non-empty string after trimming**, checking `entryLabel` and then `chipLabel`, so a present-but-empty or whitespace-only `entryLabel` no longer hides a usable `chipLabel`. That matches the legacy `JournalDataExportItem` ordering and fixes the `""` masking case reviewers flagged. Encoding is unchanged: we still write only `fullText`, so newly saved data stays normalized.

**Testing:** I’m adding a focused `Codable` decode test (or follow-up commit) that covers missing `fullText`, empty/whitespace `entryLabel` with non-empty `chipLabel`, and the happy path with `entryLabel` only — so import and store compatibility do not regress quietly. Broader load behavior remains covered by macOS CI (`grace test` / `grace ci`) on the simulator.

PR #324, opened 2026-04-14
Initialize past-statistics interval state from saved Advanced Settings

kipyin:
## Headline

Opening Advanced Settings now shows your saved past-statistics window immediately, not placeholder defaults.

## User impact

If you customized how far back past statistics look, the interval controls should match that choice the moment the screen appears. That avoids a confusing flash or moment where the UI looked like a different range or mode before it caught up.

## What changed

The past-statistics interval pickers used fixed default state until the view finished appearing and could reload from storage. An initializer was added so the same saved preference read from UserDefaults seeds the picker mode, quantity, and unit up front, aligned with the existing validation path used elsewhere on this screen.

## Verification

On a Mac with Xcode and the iOS Simulator, run `grace ci` (or `grace test` with the project’s usual destination). Open Settings → Advanced and confirm the Past statistics interval reflects your saved choice on first load without a visible mismatch.

## Risk

Medium

## Touch class

`ui-ux`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Settings/AdvancedSettingsScreen.swift
+++ b/GraceNotes/GraceNotes/Features/Settings/AdvancedSettingsScreen.swift
@@ -16,9 +16,17 @@ struct AdvancedSettingsScreen: View {
     @AppStorage(JournalAppearanceStorageKeys.todayMode)
     private var journalTodayAppearanceRaw = JournalAppearanceMode.standard.rawValue
 
-    @State private var intervalMode: PastStatisticsIntervalPickerMode = .all
-    @State private var customQuantity: Int = 4
-    @State private var customUnit: PastStatisticsIntervalUnit = .week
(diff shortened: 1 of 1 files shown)


kipyin:
**Reply to review**

We aligned Advanced Settings’ initial picker state with the same string `@AppStorage` uses for past statistics: `init` now reads the stored value through `PastStatisticsIntervalPreference.appStorageRawValue()` (same key and `""` fallback as the property wrapper) and decodes with `selection(fromAppStorage:)`, so the first paint matches persisted “all” vs custom quantity/unit without waiting for `onAppear`/`onChange`. `PastStatisticsIntervalPreference` centralizes that read and keeps decoding in one place. On this Mac I’ll sanity-check cold launch and background return with a saved custom interval (e.g. 8 weeks) and run `grace test` / `grace ci` as usual; no Linux-only verification.

PR #325, opened 2026-04-14
Guard calendar peek height against non-finite geometry

kipyin:
## Headline

Prevent bad layout math from breaking the review history calendar peek height.

## User impact

If layout ever produced NaN or infinite heights, those values could flow into the feathered calendar grid’s frame sizing and cause glitches or instability. The peek viewport now treats invalid numbers as zero height, so the calendar stays in a safe, predictable state instead of propagating broken geometry.

## What changed

The helper that clamps remaining height for the review history drilldown peek was updated so it no longer passes through non-finite values from upstream layout math. Documented that NaN and infinity collapse to zero before applying the existing non-negative clamp, keeping the calendar grid from receiving invalid frame dimensions.

## Verification

On a Mac with Xcode and the project’s dev tooling, run `grace ci` (or `grace test` for the usual simulator test pass) to confirm the app still builds and tests pass.

## Risk

Medium

## Touch class

`ui-ux`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Views/ReviewHistoryDrilldownPeekMetrics.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Views/ReviewHistoryDrilldownPeekMetrics.swift
@@ -14,7 +14,11 @@ enum ReviewHistoryDrilldownPeekMetrics {
     static let minimumViewportHeight: CGFloat = 200
 
     /// Uses at most the non-negative space remaining; never inflates above `remainingHeight`.
+    ///
+    /// Non-finite values (NaN or infinity) collapse to `0` so invalid geometry math cannot propagate into
+    /// frame sizes for the calendar grid.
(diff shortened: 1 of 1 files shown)


kipyin:
Added three unit tests in `PastDrilldownCalendarLayoutTests` so `ReviewHistoryDrilldownPeekMetrics.clampedViewportHeight(remainingHeight:)` is explicitly checked when `remainingHeight` is `NaN`, positive infinity, or negative infinity—each case must resolve to `0`. That matches the reviewer request and documents the intended contract: non-finite geometry from `GeometryReader` (or similar) must not propagate into layout as invalid frame sizes. Together with the existing finite cases, this reduces the risk of a future refactor reintroducing `max(0, x)`-style behavior that fails for `NaN` or infinities. No production code changes in this pass beyond what was already merged for the `isFinite` guard; this is test-only follow-up.

PR #327, opened 2026-04-14
Align guided journal completion reads with migration resolver

kipyin:
## Headline

Existing users no longer get misread as never finishing the guided journal when the flag was never stored.

## User impact

If the guided-journal completion flag was missing from settings, the app could treat long-time users like first-time users and surface the wrong onboarding or tutorial prompts. Reading completion through the same resolver used elsewhere restores the intended “already onboarded” behavior for upgrades and edge cases.

## What changed

The `hasCompletedGuidedJournal` getter no longer reads a bare boolean from `UserDefaults`, which defaults to false when the key is absent. It now delegates to the existing `resolvedHasCompletedGuidedJournal` helper so the first read can infer completion from legacy signals, persist a concrete value, and stay consistent with call sites that already used the static API.

The setter is unchanged: it still writes an explicit true or false when the app records completion.

## Verification

On a Mac with Xcode and the project’s `grace` CLI installed, run `grace ci` (or at least `grace test` with the repo’s default simulator destination) to confirm the app builds and automated tests pass.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Tutorial/JournalOnboardingProgress.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Tutorial/JournalOnboardingProgress.swift
@@ -39,7 +39,7 @@ final class JournalOnboardingProgress {
     }
 
     var hasCompletedGuidedJournal: Bool {
-        get { defaults.bool(forKey: JournalOnboardingStorageKeys.completedGuidedJournal) }
+        get { Self.resolvedHasCompletedGuidedJournal(using: defaults) }
         set { defaults.set(newValue, forKey: JournalOnboardingStorageKeys.completedGuidedJournal) }
     }
 


kipyin:
**Sentry — review feedback**

### PR comment (body only)

This update makes `JournalOnboardingProgress.hasCompletedGuidedJournal` use the same resolution path as `resolvedHasCompletedGuidedJournal(using:)`, so when `completedGuidedJournal` is absent but a legacy “already completed” signal is present, reads stay consistent with app launch and migration instead of falling back incorrectly. The first read may persist the resolved value, which matches how the rest of the flow materializes that key after migration. Reviewers noted a worthwhile follow-up: a unit test that seeds only the legacy flag (no `completedGuidedJournal`), then asserts the instance getter is `true` and the key is written—full app tests still run on macOS with `grace test` / `grace ci` per project docs. SwiftLint was run on the touched file; simulator validation was not run in this environment.

PR #328, opened 2026-04-14
Simplify iCloud account status fetch and log failures

kipyin:
## Headline

Faster, clearer iCloud status checks with actionable diagnostics when something goes wrong.

## User impact

Settings that depend on iCloud account state stay accurate without extra background hop. When CloudKit cannot report status, developers get a clear log line instead of a silent fallback, which makes real issues easier to find.

## What changed

The service no longer wraps the CloudKit call in a detached task. The API already suspends cooperatively, so the previous pattern added complexity without a user-facing benefit. Errors are now logged with the unified logging system, except for normal task cancellation, so noise stays low. The doc comment now describes the actual behavior instead of implying work always leaves the main actor.

## Verification

On a Mac with Xcode and the project’s grace CLI: run `grace ci` (or `grace test` with the default simulator destination from gracenotes-dev.toml). Optionally open Settings and confirm iCloud-related messaging still matches account state; if you simulate a failure, check Console for the new log under the ICloudAccountStatus category.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Settings/Services/ICloudAccountStatusService.swift
+++ b/GraceNotes/GraceNotes/Features/Settings/Services/ICloudAccountStatusService.swift
@@ -1,7 +1,13 @@
 import CloudKit
 import Foundation
+import os
 
-/// Live `CKContainer.accountStatus()` bridge; work runs off the main actor.
+private let iCloudAccountStatusLogger = Logger(
+    subsystem: "com.gracenotes.GraceNotes",
+    category: "ICloudAccountStatus"
+)
+
+/// Live `CKContainer.accountStatus()` bridge; `accountStatus()` suspends without blocking the caller.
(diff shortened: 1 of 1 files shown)


kipyin:
We aligned the CloudKit `accountStatus()` call with the earlier intent by running it inside a **utility-priority** `Task` so that work is deferred off the default high priority and cancellation still flows through the structured task (instead of a fully detached subtree). Failure logging now records the **error type**, **NSError domain**, and **code** with explicit privacy labels, so we avoid relying on `localizedDescription` for diagnostics while keeping logs usable on-device. UI behavior is unchanged: any failure still maps to `.couldNotDetermine`, and we still skip logging for `CancellationError`. On a Mac with Xcode, it’s worth a quick pass in Settings (and any tour path that refreshes status) plus rapid navigation to confirm there’s no noticeable hitch or noisy duplicate logs—that matches what automated tests don’t fully cover.

PR #329, opened 2026-04-14
Fix demo seed collision between anchored history and rolling week

kipyin:
## Headline

Demo journal seeding no longer overwrites a week row when a fixed history date lands on the same day.

## User impact

In the Demo build, two seed rows that share the same calendar day could fight during upsert, so the last write could replace the intended “this week” sample with the anchored history sample (or the other way around). The seeder now also uses one captured “now” for the skip check and for building rows so “today” does not shift mid-run.

## What changed

The demo seed version was bumped so existing Demo installs pick up the new behavior. Seeding captures a single current time up front and passes it into the “should we seed?” check, and that path uses the same instant for the start-of-day “today” test instead of reading the clock again.

When assembling demo rows, the code builds the rolling seven-day set first, then adds the fixed December 2025 history row only if that anchored day is not already represented in the week. If the calendar would collide, the anchored row is skipped so upserts cannot clobber the week’s intended story for that day.

## Verification

On a Mac with Xcode and simulators installed, run `grace ci` from the repo root. For a manual spot-check, launch the GraceNotes (Demo) scheme and confirm the demo journal still looks coherent around the current week and, when the anchored day does not overlap the week, that the separate past-year sample still appears.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Data/Persistence/DemoDataSeeder.swift
+++ b/GraceNotes/GraceNotes/Data/Persistence/DemoDataSeeder.swift
@@ -4,17 +4,17 @@ import SwiftData
 
 @MainActor
 enum DemoDataSeeder {
-    private static let seedVersion = 5
+    private static let seedVersion = 6
     private static let seedVersionKey = "demoDataSeedVersion"
 
     static func seedIfNeeded(context: ModelContext, calendar: Calendar = .current) {
         let seedTrace = PerformanceTrace.begin("DemoDataSeeder.seedIfNeeded")
-        guard shouldSeed(context: context, calendar: calendar) else {
+        let now = Date.now
+        guard shouldSeed(context: context, calendar: calendar, now: now) else {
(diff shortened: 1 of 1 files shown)


kipyin:
Thanks for the careful read. I replaced the fixed December 2025 anchor with a historical row on the calendar day **before** the oldest day in the rolling week (`today` minus seven days, normalized to start-of-day), so the anchor and the seven demo days never share a day and upsert-by-day cannot clobber the week. I added a short comment in `makeSeedEntries` explaining that guarantee. `shouldSeed` and seeding now use the same `now` snapshot so “today” does not drift between the guard and `makeSeedEntries`. I did not add a separate test target for `#if USE_DEMO_DATABASE` seeding in this pass; if we want automated coverage next, extracting the date math into a small testable helper would be the lightest path. Manual check on macOS remains: open **GraceNotes (Demo)** and confirm the week plus one older row look right after a version bump.

PR #330, opened 2026-04-14
Centralize clearing of journal tutorial UserDefaults keys

kipyin:
## Headline

Tutorial hints and celebrations reset from one authoritative list of storage keys.

## User impact

Resets and test cleanups now reliably wipe every current and legacy tutorial flag, so hints and milestone celebrations do not reappear because an old key was missed. This keeps first-run guidance predictable after a reset and makes automated UI tests less flaky.

## What changed

Tutorial storage keys were consolidated into two explicit lists: one for the current UserDefaults names and one for the legacy spellings kept for migration. A new helper on the keys type removes all of those entries in one place. The journal tutorial progress reset path now calls that helper instead of maintaining a duplicate list and hard-coded legacy strings inline, so future key changes only need to be updated in one spot.

## Verification

On a Mac with Xcode and the project’s grace CLI installed, run `grace ci` (or `grace test` with the repo’s default simulator destination) to confirm the app still builds and tests pass. Optionally reset tutorial state via the code path that calls reset and confirm Today tutorial hints and celebrations behave as a fresh install.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Tutorial/JournalTutorialProgress.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Tutorial/JournalTutorialProgress.swift
@@ -11,6 +11,14 @@ enum JournalTutorialStorageKeys {
     /// First time all fifteen Entries were filled (Bloom).
     static let celebratedFirstBloom = "journalTutorial.celebratedFirstBloom"
 
+    private static let currentKeys: [String] = [
+        dismissedSproutGuidance,
+        dismissedBloomGuidance,
+        celebratedFirstSprout,
+        celebratedFirstLeaf,
+        celebratedFirstBloom
+    ]
+
     /// Copies legacy keys into the keys above, then removes legacy entries.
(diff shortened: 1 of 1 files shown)


kipyin:
Thanks for the deep review. Tutorial reset now clears keys through a single `currentKeys` list plus the legacy key list, so `resetAll` and `-reset-journal-tutorial` stay aligned with the canonical string constants instead of duplicating an ad hoc array in `JournalTutorialProgress`. I added a short note next to those constants so anyone who introduces a new `UserDefaults` key is reminded to append it to `currentKeys` and to extend migration when replacing a legacy spelling. I agree that forgetting `currentKeys` would still cause drift; a small unit test that the cleared key set matches what the progress accessors use would be a good hardening follow-up if we want to lock that in. Residual risks you called out (pre-legacy unknown keys, `UserDefaults` concurrency) match my understanding and were not part of this change.

PR #331, opened 2026-04-14
Fix CompletionInfoCard VoiceOver and replace AnyView card surface

kipyin:
## Headline

Cleaner accessibility for the journal completion card and more predictable SwiftUI rendering for the morph animation.

## User impact

VoiceOver no longer risks repeating or overriding the card’s description when the text is already exposed as child content, so the completion message is easier to hear correctly. The card’s animated surface is also expressed without type erasure, which helps SwiftUI stay efficient and stable during the completion morph.

## What changed

Removed the container-level accessibility label on the completion info card because the view is already configured to expose its children to assistive technologies, which avoids redundant or conflicting announcements alongside the visible description text.

Rebuilt the card background so the rounded rectangle and bloom styling live in one shared subview, and applied the matched geometry effect only when morphing is enabled and reduced motion is off. That replaces the previous AnyView-based branching with ordinary conditional views, which is easier for the framework to reason about and aligns with typical SwiftUI style.

## Verification

On a Mac with Xcode and the repo’s grace CLI, run `grace ci` (or at least `swiftlint lint` plus a simulator build via your usual grace workflow) and exercise the journal completion flow with VoiceOver on: confirm the card’s message is announced once clearly, and with Reduce Motion off confirm the morph animation still looks correct.

## Risk

Medium

## Touch class

`ui-ux`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Views/CompletionInfoCard.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Views/CompletionInfoCard.swift
@@ -30,14 +30,30 @@ struct CompletionInfoCard: View {
         )
         .shadow(color: cardTintColor.opacity(reduceTransparency ? 0.18 : 0.24), radius: 8, x: 0, y: 4)
         .accessibilityElement(children: .contain)
-        .accessibilityLabel(Text(badgeInfo.description))
         .onAppear {
             animateEntry()
         }
     }
 
-    private var cardSurface: AnyView {
-        let base = RoundedRectangle(cornerRadius: AppTheme.cornerRadiusMedium)
+    private var cardSurface: some View {
+        Group {
+            if showMorph && !reduceMotion {
+                roundedRectangleSurface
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

This update removes the extra container-level accessibility label on `CompletionInfoCard` so VoiceOver reads the card’s message once clearly, while keeping `accessibilityElement(children: .contain)` so the inner text stays the natural focus. The morph animation path now uses a typed `Group` and shared `roundedRectangleSurface` instead of `AnyView` around `matchedGeometryEffect`, which keeps view identity clearer and avoids unnecessary type erasure. Today’s presentation uses `showMorph: false`, so the morph branch is not the default path; I’ll still sanity-check VoiceOver over the dimmed scrim in `DateSectionView` on Simulator or a device. I agree there’s no automated coverage for labels or geometry effects here—manual QA remains the practical guardrail for accessibility and animation regressions.

PR #332, opened 2026-04-14
Treat CloudKit and URL cancellations as non-errors in iCloud status fetch

kipyin:
## Headline

Stop logging benign cancellations when checking iCloud account status so diagnostics stay trustworthy.

## User impact

When the app cancels or CloudKit reports a normal cancellation, you should not see misleading “failed to fetch” error noise in logs. Real failures still surface clearly, which makes support and debugging more reliable.

## What changed

The iCloud account status helper only skipped logging for Swift’s `CancellationError`. CloudKit and the networking stack often surface cancellation differently (for example operation cancelled or URL cancelled), so those were incorrectly treated as failures and logged as errors.

The change adds a small helper that treats those cancellation-shaped errors the same as `CancellationError`: we still return “could not determine” for the UI path, but we no longer log an error for ordinary cancellation.

## Verification

On a Mac with the repo and Xcode, run `grace ci` (or `grace test` with your usual simulator destination) to confirm the project still builds and tests pass. Optionally exercise Settings flows that touch iCloud status and confirm logs stay quiet when work is cancelled.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Settings/Services/ICloudAccountStatusService.swift
+++ b/GraceNotes/GraceNotes/Features/Settings/Services/ICloudAccountStatusService.swift
@@ -21,7 +21,7 @@ final class ICloudAccountStatusService: ICloudAccountStatusProviding {
             let status = try await container.accountStatus()
             return ICloudAccountBucket(status)
         } catch {
-            if !(error is CancellationError) {
+            if !isCancellationLike(error) {
                 let detail = error.localizedDescription
                 iCloudAccountStatusLogger.error(
                     "Failed to fetch iCloud account status. \(detail, privacy: .public)"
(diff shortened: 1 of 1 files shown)


kipyin:
Updated the doc comment on `isCancellationLike` to say **`CKError.Code.operationCancelled`** so it matches how we actually detect cancellation (`ckError.code == .operationCancelled` and NSError domain/code), and so it lines up with `URLError.cancelled` in the same note.

Cancellation handling now **unwraps underlying errors**: we walk **`NSUnderlyingErrorKey`** and **`NSUnderlyingErrorsKey`** recursively (with a depth limit) so wrapped `CKError`, `CancellationError`, or URL cancellation still count as cancellation-like and **do not** hit the error log or get misclassified as an ordinary failure.

Added **`ICloudAccountStatusCancellationTests`** in **GraceNotesTests** to cover direct cancellation errors, NSError bridging, underlying-error chains, and a non-cancellation case so this behavior does not regress silently.

**Verification:** `grace test` / `grace ci` on macOS (XCTest and CloudKit types are not available on Linux agents).

PR #333, opened 2026-04-14
Fix journal history month grouping when month keys cannot be built

kipyin:
## Headline

Journal history now keeps all entries in the correct month section instead of splitting one month into many.

## User impact

Entries were sometimes shown under lots of tiny “month” rows instead of a single month, which made history hard to scan and felt broken. Grouping now uses a stable start-of-month key so months stay together reliably.

## What changed

Month grouping no longer falls back to each entry’s exact timestamp when building the month bucket key. A small helper picks the start of the calendar month when possible, normalizes year/month to the first day with time cleared if needed, and only then uses a last-resort day-level anchor—so failed date construction cannot scatter same-month entries across many groups.

## Verification

On a Mac with the repo’s dev tooling, run `grace ci` (or `grace test` with your usual simulator destination from `gracenotes-dev.toml`) and spot-check Journal history: entries for a given calendar month should appear under one month heading, including around calendar or data edge cases.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/HistoryEntryGrouping.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/HistoryEntryGrouping.swift
@@ -7,12 +7,29 @@ enum HistoryEntryGrouping {
         calendar: Calendar
     ) -> [(key: Date, entries: [Journal])] {
         let grouped = Dictionary(grouping: entries) { entry -> Date in
-            let components = calendar.dateComponents([.year, .month], from: entry.entryDate)
-            return calendar.date(from: components) ?? entry.entryDate
+            monthKey(for: entry.entryDate, calendar: calendar)
         }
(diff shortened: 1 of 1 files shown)


kipyin:
**What changed**

Clarified the `monthKey` documentation so it matches the real three-step behavior: we aim for the start of the calendar month, but when month-level normalization cannot be produced we fall back to the start of the day so entries still bucket predictably instead of fragmenting by raw timestamps. Added a regression test that drives the failure path (a calendar/date setup where month-key construction returns `nil`) and asserts entries in the same calendar month still land in a single month section. This directly addresses the review concern that the old comment overstated guarantees and that the new branches were untested.

**Verification**

`swiftlint lint` from the repo root. On macOS, run the updated `HistoryEntryGroupingTests` through Xcode or `grace test` to execute the XCTest suite.

PR #334, opened 2026-04-14
Fix month banners when a week spans two months

kipyin:
## Headline

The review history drill-down calendar now labels both months when a single week row crosses a month boundary.

## User impact

Week rows that include days from two months used to pick only the first dated cell for the month banner, so the calendar could skip or misrepresent the transition between months. Users scanning review history now see the correct month heading(s) for that week, which improves clarity and supports VoiceOver and scrolling that rely on those structural rows.

## What changed

Month banner rows are no longer derived from only the first non-empty day in each week. The layout walks week cells in order, collects each distinct year-month in that order, and inserts a banner for each new month key that has not already been shown, so a week that straddles two months can produce two banners before the week row. Week row identity still uses the first real day in the row, with a small refactor to avoid repeating the compact-map pass.

## Verification

On a Mac with Xcode and the iOS Simulator, run `grace ci` (or `grace test` with the project’s usual destination) and manually open the review history drill-down calendar around a month boundary to confirm both month labels appear when a week spans two months.

## Risk

Medium

## Touch class

`ui-ux`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Views/ReviewHistoryDrilldownCalendar.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Views/ReviewHistoryDrilldownCalendar.swift
@@ -144,6 +144,25 @@ enum ReviewHistoryDrilldownCalendarLayout {
         return weeks
     }
 
+    /// Month keys in cell order (year-month). Weeks spanning two months yield two specs.
+    private static func orderedMonthBannerSpecs(
+        for week: [Date?],
+        calendar: Calendar
+    ) -> [(key: String, monthDate: Date)] {
+        var seen = Set<String>()
+        var specs: [(String, Date)] = []
+        for cell in week {
+            guard let date = cell else { continue }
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

Thanks for the thorough review. This change fixes the past-drilldown calendar so a single padded week row that includes real days from two different calendar months gets **two** month banners in cell order (January then February, for example), instead of only the first month’s label. We derive those banners from `orderedMonthBannerSpecs`, which walks the week’s non-nil cells left to right and collects distinct year–month keys in order, and we use `lastBannerMonthKey` when walking weeks so we do not repeat a banner when the same month continues from the previous row. The existing Jan–Mar range test still checks that multiple months in the viewport produce the right number of banners; we agree a follow-up test that asserts **two banners immediately before one `.week` row** for a split week would lock that invariant more directly. Those tests run on **macOS** in CI via `grace test` / `grace ci` (not on Linux), so adding that case would still be worthwhile hardening.

PR #335, opened 2026-04-14
Ignore completion highlight in rhythm scroll pin identity

kipyin:
## Headline

Weekly review rhythm keeps your horizontal scroll position when only completion highlights refresh.

## User impact

Previously, when the app recomputed which growth-stage row was highlighted for a day, the scroll pin treated that as a layout change and could reset horizontal scroll. That felt jumpy and could pull you away from where you were reading. Pin identity now tracks only what actually changes column count and horizontal extent.

## What changed

The scroll pin identity for the reflection rhythm strip used full struct equality, so any change to per-day completion staging (strongest completion level) looked like a new dataset and cleared trailing-edge scroll pinning. Custom equality now compares week start and, for each day in order, the date plus whether there is reflective activity and a persisted entry—matching what affects visible columns—while ignoring the completion highlight field. Documentation explains why that field is excluded.

## Verification

On a Mac with Xcode and the Grace Notes dev CLI, run `grace ci` (or `grace test` with the project’s default simulator destination) to lint and build; exercise the weekly review rhythm strip: scroll horizontally, trigger a path that recomputes completion highlights without adding/removing columns, and confirm scroll position stays pinned.

## Risk

Medium

## Touch class

`ui-ux`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Views/ReviewRhythmScrollPinIdentity.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Views/ReviewRhythmScrollPinIdentity.swift
@@ -4,7 +4,19 @@ import Foundation
 ///
 /// Used to reset trailing-edge scroll pinning after a real dataset change, without clearing user scroll when
 /// `ReviewInsights` is regenerated for the same week with the same per-day rhythm payload (#131).
+///
+/// Equality ignores ``ReviewDayActivity/strongestCompletionLevel``: it only affects which growth-stage row is
+/// highlighted, not column count or horizontal extent, so recomputing completion staging must not reset pin state.
 struct ReviewRhythmScrollPinIdentity: Equatable {
     let weekStart: Date
     let days: [ReviewDayActivity]
+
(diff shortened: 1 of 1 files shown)


kipyin:
**Response to review**

We tightened `ReviewRhythmScrollPinIdentity` so equality only depends on the week anchor, the number of rhythm columns, and each column’s calendar date (in order). That matches how `ReviewSummaryCard` lays out the horizontal rhythm strip: width is driven by how many days there are and which dates they are, not by per-day content like reflective activity, persisted entry, or completion highlight level—so those fields no longer reset trailing-edge pinning when insights refresh without a layout change. The struct’s documentation now states that explicitly and notes that any future `ReviewDayActivity` field that affects column count or width should be folded into this comparison. We added unit tests that two identities stay equal when only `hasReflectiveActivity` / `hasPersistedEntry` differ and when only `strongestCompletionLevel` differs, plus existing negative cases for week and date changes. A dedicated UI test for scroll position stability on completion-only updates is still a reasonable follow-up if we want end-to-end regression coverage beyond these equality tests.

PR #336, opened 2026-04-14
Clamp reminder journal-day gap to non-negative values

kipyin:
## Headline

Daily journal reminder copy no longer treats future-dated entries as a negative “days since” gap.

## User impact

If your latest meaningful entry was dated after today, the app could compute a negative day gap, which could skew which reminder message variant you get (for example, lapse-style copy). The gap is now treated as zero days in that edge case, so messaging stays aligned with “today.”

## What changed

The calendar-day gap helper that feeds reminder body selection used the raw day count between the last meaningful journal day and the start of today. When the last meaningful day is after today, that difference can be negative or the day component can be missing. The update clamps the result to zero at minimum, treats a missing day as zero, and documents that the returned gap is never negative.

## Verification

On a Mac with Xcode and the iOS Simulator, run `grace ci` (or `grace test` using the project’s configured simulator destination). If there are existing tests around `ReminderNotificationBodySelector`, run those or the full test target to confirm behavior.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Services/Reminders/ReminderNotificationBodySelector.swift
+++ b/GraceNotes/GraceNotes/Services/Reminders/ReminderNotificationBodySelector.swift
@@ -73,6 +73,8 @@ enum ReminderNotificationBodySelector {
 
     /// Calendar-day gap from the last meaningful journal day to ``todayStart``.
     /// Returns `nil` when there was never meaningful content.
+    /// The result is always non-negative: future-dated entries (``lastDay`` after ``todayStart``)
+    /// yield `0` instead of a negative component.
     static func calendarDayGapSinceLastMeaningfulEntry(
         entries: [Journal],
         todayStart: Date,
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

Thanks for the thorough review. This change tightens **meaningful-day gap** so we never treat a **future** last meaningful calendar day as a large positive gap: we **clamp** the calendar day difference to **at least 0**, document **future-dated entries as gap 0**, and keep lapse logic (`gapDays >= 3`) consistent with that contract—fixing the case where a **negative** gap could wrongly suppress a lapse. I agree a small **XCTest** with `lastMeaningful` after `todayStart` would pin that behavior; if you want that in this PR, I can add it. The **gap 0** semantics for “tomorrow” stay intentional for notifications; **DST** behavior remains the same as before because we still rely on **Calendar**’s **day** component.

PR #337, opened 2026-04-14
Normalize export history detail for single-line display

kipyin:
## Headline

Export history and VoiceOver now show trimmed, one-line detail instead of raw multi-line or whitespace-only strings.

## User impact

History rows and the plain accessibility label stay on a single line and read cleanly. Whitespace-only stored detail no longer appears as a confusing empty fragment, and embedded line breaks no longer break the layout or speech flow.

## What changed

Export history previously treated any non-empty stored detail as visible, which could surface padding-only text or multi-line blobs in Settings and in the shared plain label used for VoiceOver.

Detail is now normalized: leading and trailing whitespace are stripped, runs of line breaks become spaces so the string fits one line, and if nothing meaningful remains the detail is omitted entirely. Documentation on the detail field was updated to match this rule.

## Verification

On a Mac with the repo and `grace` installed, run `grace ci`. Optionally open Settings → Data & Privacy, find import/export export history, and confirm entries with messy stored detail read as a single line with sensible spacing.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Settings/ImportExportTechnicalDetailFormatting.swift
+++ b/GraceNotes/GraceNotes/Features/Settings/ImportExportTechnicalDetailFormatting.swift
@@ -3,7 +3,7 @@ import Foundation
 struct ExportHistoryLineParts {
     let kindLabel: String
     let statusLabel: String
-    /// Non-nil when `entry.detail` is non-empty (same rule as on-screen history).
+    /// Non-nil when `entry.detail` has visible content after normalization (trim; line breaks flattened).
     let detail: String?
 }
 
@@ -34,12 +34,7 @@ enum ImportExportTechnicalDetailFormatting {
         } else {
             statusLabel = String(localized: "settings.dataPrivacy.importExport.history.status.failed")
         }
-        let detail: String?
-        if let raw = entry.detail, !raw.isEmpty {
-            detail = raw
-        } else {
-            detail = nil
-        }
+        let detail = entry.detail.flatMap { Self.normalizedExportHistoryDetail($0) }
         return ExportHistoryLineParts(kindLabel: kindLabel, statusLabel: statusLabel, detail: detail)
     }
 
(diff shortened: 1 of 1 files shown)


kipyin:
Thanks for the thorough review. Here is a concise PR comment body the author can post:

---

Added `XCTest` coverage for export history detail normalization so multiline strings become a single spaced line, whitespace-only detail becomes absent (`parts.detail == nil`) with a two-segment plain label, and leading/trailing newlines still preserve inner text. Updated `normalizedExportHistoryDetail` to split on newlines, trim each segment, drop empty segments, and join with a single space so inputs like `foo \n bar` or blank lines between words do not produce double spaces in the UI or VoiceOver string. That locks the rules reviewers called out and should let CI on macOS (`grace test` / `grace ci`) catch future regressions in `exportHistoryLineParts` and `exportHistoryPlainLabel`.

---

Let me tighten to 2-6 sentences as requested - the user said 2-6 sentences. Let me count:

1. Added XCTest coverage...
2. Updated normalizedExportHistoryDetail...
3. That locks...

That's 3 sentences. Good.

Actually re-read - they want ONLY the comment body. I'll refine for flow and ensure it's exactly what they need - public GitHub PR comment responding to review.

Version:

Thanks for the feedback. I added focused unit tests in `ImportExportDetailFormattingTests` that assert multiline detail becomes one line with single spaces, whitespace-only detail yields no detail segment and a two-part plain label, and edge cases like padded newlines or blank lines do not introduce double spaces. I also changed `normalizedExportHistoryDetail` to trim each newline-delimited segment and omit empty segments before joining, matching the “trimmed, one-line detail” goal for both on-screen text and `exportHistoryPlainLabel`. Full `GraceNotesTests` still need to run on macOS in CI after merge for compile and runtime confidence.

That's good. 3 sentences.

User said: "Explain what you changed and why" - good.

"No greeting or sign-off" - I had "Thanks for the feedback" - that might count as greeting. Remove it.

Final:

I added focused unit tests in `ImportExportDetailFormattingTests` that assert multiline detail becomes one line with single spaces, whitespace-only detail yields no detail segment and a two-part plain label, and messy inputs (e.g. spaces around newlines or blank lines between words) do not produce double spaces in the composed string. `normalizedExportHistoryDetail` now trims each segment after splitting on newlines, drops whitespace-only segments, and joins with a single space so the visible history line and VoiceOver label stay consistently single-line and trimmed. Full test execution remains on macOS via the project’s `grace` CLI in CI.

4 sentences. Good.

File names in backticks is fine markdown.

I'll output ONLY the comment body as requested.
`</think>`
I added focused unit tests in `ImportExportDetailFormattingTests` that cover multiline detail collapsing to one line with single spaces, whitespace-only detail becoming absent (`parts.detail == nil`) with a two-segment `exportHistoryPlainLabel`, and messy inputs (spaces around newlines, blank lines between words, leading/trailing newlines) so we do not regress double spaces or inner content. `normalizedExportHistoryDetail` now splits on newlines, trims each segment, drops empty segments, and joins with a single space, which matches the intended trimmed, one-line behavior for both on-screen text and accessibility. Validation of the full `GraceNotesTests` target should continue to run on macOS through the repo’s `grace test` / `grace ci` path in CI.

PR #339, opened 2026-04-14
Fix weekly review theme tie-break order for first-seen themes

kipyin:
## Headline

Most recurring themes now break ties using each theme’s true first appearance, not per journal surface.

## User impact

When two themes show up equally often in your weekly review, the list order could feel arbitrary or shuffle relative to how you actually wrote them. Ordering is now steadier and matches the sequence themes are first detected across your entries.

## What changed

Theme aggregation used a single “first seen” sequence number for every new theme that appeared on the same section of an entry, then advanced that counter only after the whole section was processed. That meant co-occurring themes could look tied even when the extractor had a clear order, so tie-breaking could lean on label sorting instead of discovery order. The builder now advances the sequence only when a canonical theme is seen for the first time, so each theme gets a distinct first-seen rank that follows the actual scan order.

## Verification

On a Mac with the repo’s dev toolchain, run `grace ci` (or `grace test` with the project’s default simulator destination) to confirm the GraceNotes scheme still builds and tests pass.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Services/WeeklyReviewAggregatesBuilder+ThemeSections.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Services/WeeklyReviewAggregatesBuilder+ThemeSections.swift
@@ -52,6 +52,7 @@ extension WeeklyReviewAggregatesBuilder {
                     .compactMap { _, candidates in candidates.max(by: { $0.score < $1.score }) }
 
                 for concept in uniqueConcepts {
+                    let isFirstAppearance = map[concept.canonicalConcept] == nil
                     var accumulator = map[concept.canonicalConcept] ?? DistilledThemeAccumulator(
                         canonicalConcept: concept.canonicalConcept,
                         displayLabel: concept.displayLabel,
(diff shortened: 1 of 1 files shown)


kipyin:
We replaced `Dictionary(grouping:)` with a single pass over the distilled concepts that keeps the strongest candidate per canonical label while preserving the extractor’s emitted order, so `firstSeenOrder` tracks real debut order within a surface instead of whatever order grouped dictionary keys happened to iterate. We also advance `sequence` only when a canonical theme is first inserted into the accumulator, so themes that show up together in one scan are no longer forced into an artificial tie on `firstSeenOrder`. For ties on total count and day count, most recurring ordering now uses that stable debut order before falling back to string comparison. A new unit test builds two entries that repeat the same multi-concept gratitude line so Walking and Rest tie on counts and days, then asserts list order matches whichever canonical appeared first in `distillConcepts` for that phrase—locking the regression reviewers called out. Validation on macOS is still `grace ci` / `grace test` as in the PR description.

PR #340, opened 2026-04-14
Defer reminder disable until after in-flight reminder operations

kipyin:
## Headline

Turning reminders off now completes even if you toggle while a schedule change is still running.

## User impact

If someone turned off daily reminders while the app was busy rescheduling or saving the time, the old flow could exit early and leave reminders enabled. That mismatch is confusing and can mean notifications keep firing when the user expects them to stop.

## What changed

Reminder settings used to bail out of disable when another reminder task had already set the working state, so a fast “off” tap during reschedule could be dropped.

The model now cancels any pending debounced reschedule, clears coalesced save flags, and remembers a disable request when work is in flight. After that work finishes, it runs the deferred disable so notifications are actually turned off and the live status refreshes. The same drain runs after enable and after chained reschedule work so the off path stays consistent.

## Verification

On a Mac with Xcode and the iOS Simulator, exercise Settings → reminders: enable, change time rapidly, then toggle off while a save is in progress; confirm reminders end up off and the UI matches. Run `grace ci` (or `grace test` with the project’s usual destination) to confirm the app still builds and tests pass.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Settings/ReminderSettingsFlowModel.swift
+++ b/GraceNotes/GraceNotes/Features/Settings/ReminderSettingsFlowModel.swift
@@ -15,6 +15,8 @@ final class ReminderSettingsFlowModel: ObservableObject {
     private let userDefaults: UserDefaults
     private var pendingRescheduleTask: Task<Void, Never>?
     private var pendingRescheduleAfterCurrentSave = false
+    /// User turned reminders off while another reminder operation held `isWorking`.
+    private var pendingDisableAfterCurrentWork = false
     private var hasLoadedLiveStatus = false
 
     init(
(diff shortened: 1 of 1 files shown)


kipyin:
Addressed the deferred-disable race so reminder state matches what the user last asked for. **`enableReminders()`** clears **`pendingDisableAfterCurrentWork`** up front so turning reminders back on during an in-flight reschedule/enable isn’t undone when the busy window ends and drains run. **`disableReminders()`** also clears the pending flag after a direct disable attempt finishes, so extra disable calls that stack up while **`disableDailyReminder`** is awaiting don’t leave a stale flag for a later enable/reschedule drain. Added unit tests that hold reschedule/ disable with **`RescheduleTestGate`**: deferred disable after reschedule, off-then-on mid-reschedule (no surprise disable after re-enable), and redundant disable during an in-flight disable.

PR #341, opened 2026-04-14
Stack journal onboarding actions at large Dynamic Type sizes

kipyin:
## Headline

Onboarding suggestion buttons stay readable when text is enlarged.

## User impact

People who use larger text (including accessibility sizes) were stuck with two side-by-side buttons that could feel cramped or hard to tap. The card now lays out actions in a single column at those sizes so labels and targets stay clear and reliable.

## What changed

The journal onboarding suggestion card still shows the same primary and secondary actions, but it now watches the system Dynamic Type size. At default sizes, the two buttons stay in one row as before. When Dynamic Type is in an accessibility range, the buttons switch to a vertical stack so each action can use the full width and breathe with longer labels. The button definitions were pulled into small private views so the horizontal and vertical layouts share one implementation without duplicating styling.

## Verification

On a Mac with Xcode and the iOS Simulator, run `grace ci` (or build and run the GraceNotes scheme). In the journal onboarding suggestion UI, toggle larger accessibility text sizes in Settings (or the simulator’s Environment overrides) and confirm the two actions stack vertically; at default text size, confirm they remain side by side.

## Risk

Medium

## Touch class

`ui-ux`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Tutorial/JournalOnboardingSuggestionView.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Tutorial/JournalOnboardingSuggestionView.swift
@@ -2,6 +2,7 @@ import SwiftUI
 
 struct JournalOnboardingSuggestionView: View {
     @Environment(\.todayJournalPalette) private var palette
+    @Environment(\.dynamicTypeSize) private var dynamicTypeSize
     let title: String
     let message: String
     let primaryActionTitle: String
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

No substantive code changes in this pass; the branch is unchanged from the last push. I did not address new review feedback because none was provided in the digest. If reviewers have follow-up items or want adjustments to scope, I can take those in the next iteration.

PR #342, opened 2026-04-14
Debounce iCloud sync activity writes to UserDefaults

kipyin:
## Headline

Settings’ last-sync indicator stays accurate while avoiding redundant disk writes during busy sync bursts.

## User impact

When iCloud pushes many remote changes in a short time, the app no longer writes the persisted timestamp on every single notification. That cuts unnecessary I/O and reduces wear on storage while the “last activity” signal users see in Settings remains up to date for practical purposes.

## What changed

The iCloud sync activity model still updates the in-memory “last remote change” time whenever a remote-store notification fires, so the UI can reflect fresh activity. Persisting that timestamp to UserDefaults is now coalesced: repeated notifications in a burst schedule a single main-queue flush instead of writing on every event. A small flag prevents stacking duplicate pending flushes until the scheduled write runs.

## Verification

On a Mac with Xcode and the repo’s `grace` CLI, run `grace ci` (or `grace test` with the project’s usual destination) to confirm the GraceNotes scheme still builds and tests pass; optionally exercise Settings with iCloud sync and confirm the last-activity line still updates without excessive churn under rapid sync.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Settings/Services/ICloudSyncActivityModel.swift
+++ b/GraceNotes/GraceNotes/Features/Settings/Services/ICloudSyncActivityModel.swift
@@ -10,6 +10,8 @@ final class ICloudSyncActivityModel: ObservableObject {
     @Published private(set) var lastRemoteChangeAt: Date?
 
     private var observerToken: NSObjectProtocol?
+    /// Limits disk writes when many remote-change notifications arrive in a short burst (e.g. sync).
+    private var pendingUserDefaultsPersist = false
 
     init() {
         let raw = UserDefaults.standard.double(forKey: Self.persistedTimestampKey)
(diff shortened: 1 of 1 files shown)


kipyin:
We coalesce repeated `NSPersistentStoreRemoteChange` handling so we do not write to `UserDefaults` on every notification during heavy sync bursts, which cuts redundant disk I/O while still updating `lastRemoteChangeAt` in memory each time. The deferred persist runs in a `Task { @MainActor in … }` so the write stays on the `MainActor` and matches stricter concurrency expectations. We call `persistLastRemoteChangeIfNeeded()` on **did enter background** and **will terminate** so disk is flushed when the app is leaving or exiting, reducing the window where an abrupt stop could leave an older timestamp on disk than the latest in-memory value. Tests now include firing several remote-change notifications in a row and, after async waits, asserting `UserDefaults` holds the same timestamp as the model’s latest `lastRemoteChangeAt`, locking in the “last write wins” behavior after coalescing.

PR #343, opened 2026-04-14
Harden ShareCardStyle hex colors and simplify completion chip text

kipyin:
## Headline

Share cards get more predictable colors from hex values, with one less place for style tokens to drift.

## User impact

Exported journal share cards should look the same or more consistent when colors are built from hex-style literals. The completion chip label color stays aligned with the footer “muted ink” you already use elsewhere for each style, without duplicate definitions that could get out of sync later.

## What changed

The private hex-to-color helper now treats inputs as 24-bit RGB and ignores any high bits, so alpha-prefixed style literals still map to the same sRGB triplet instead of shifting channels in surprising ways. Colors are created explicitly in the sRGB color space for predictable rendering. Completion chip text color no longer repeats a per-style switch; it reuses the existing footer ink color, which matched those values already—so behavior stays the same with simpler maintenance.

## Verification

On a Mac with Xcode and the project’s grace CLI installed, run `grace ci` from the repo root (or `grace test` with the usual simulator destination) to lint and build the GraceNotes scheme.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Services/ShareCardStyle.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Services/ShareCardStyle.swift
@@ -1,11 +1,13 @@
 import SwiftUI
 
 private extension Color {
+    /// 24-bit `RRGGBB`. High bits are masked so `AARRGGBB`-style literals still resolve to the same sRGB triplet.
     init(hex: UInt) {
-        let red = Double((hex >> 16) & 0xFF) / 255
-        let green = Double((hex >> 8) & 0xFF) / 255
-        let blue = Double(hex & 0xFF) / 255
-        self.init(red: red, green: green, blue: blue)
+        let rgb = hex & 0xFFFFFF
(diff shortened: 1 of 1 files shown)


kipyin:
We pulled hex parsing into a single **`Color.init(hex:)`** in `Color+Hex.swift` (24-bit mask and explicit sRGB), removed the duplicate private `Color` hex helpers from the share-card files, and matched **`Theme`’s `UIColor(hex:)`** to the same masking so adaptive and share-card colors stay consistent. **`completionChipTextColor`** now goes through **`footerInk`** so chip text stays in sync with footer ink with no change for current palettes; we can split them later if product wants a different rule. We agree a quick **visual pass on exported share cards** (light content, wide gamut if you have it) is the right check here—**`grace test` / `grace ci`** on a Mac covers build and tests, not bitmap fidelity. Thanks for the non-blocking note on future drift; the codebase now has one SwiftUI source of truth plus the aligned UIKit path where adaptive colors need it.

PR #344, opened 2026-04-14
Skip weekly insight candidate work when the week is sparse

kipyin:
## Headline

Weekly review no longer builds pattern insights for sparse weeks only to throw them away.

## User impact

Sparse weeks still get the same gentle fallback messaging you expect, but the app avoids unnecessary work and the flow matches the product rule: pattern-based insights are not shown when reflection is thin.

## What changed

Previously, `buildCandidates` assembled every scored insight (completion streaks, recurring people/themes, gaps, continuity) and only then checked whether the week was sparse, returning an empty list and discarding that work. The sparse-week guard now runs first and returns no candidates immediately, so we never compute insights that would not be used. Behavior for callers is unchanged: an empty candidate list still leads to the existing fallback path.

## Verification

On a Mac with Xcode, run `grace ci` (or `grace test` with the project’s default simulator destination) to confirm the GraceNotes scheme still builds and tests pass. Optionally exercise the weekly review with a sparse week of journaling and confirm fallback insights still appear as before.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Services/WeeklyInsightCandidateBuilder.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Services/WeeklyInsightCandidateBuilder.swift
@@ -13,6 +13,10 @@ struct WeeklyInsightCandidateBuilder {
     }
 
     func buildCandidates(inputs: CandidateInputs) -> [InsightCandidate] {
+        if isSparseWeek(entries: inputs.entries, reflectionDayCount: inputs.currentDayCount) {
+            return []
+        }
+
         var candidates: [InsightCandidate] = []
 
         if let fullCompletion = fullCompletionCandidate(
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

**What changed**  
We moved the sparse-week check to the **start** of `WeeklyInsightCandidateBuilder.buildCandidates` so we return early when a week is sparse instead of building candidates and then throwing them away. **Why**  
For sparse weeks the result is still an empty candidate list; for non-sparse weeks we still run the full candidate build. The engine still passes that list through `selectInsights`, so when nothing is selected the same fallback behavior applies. We did **not** change `isSparseWeek` or any heuristics—only evaluation order. **Verification**  
`grace test` / `grace ci` on macOS (per project workflow) is the right check; behavior should match existing `WeeklyInsightRuleEngine` tests as the contract.

PR #345, opened 2026-04-14
Align Review insights refresh key with a single reference date

kipyin:
## Headline

Past tab insights now compute week boundaries and history windows from one shared moment.

## User impact

When the clock or data changes while insights load, week and “past window” math can disagree if different `Date()` values are mixed in. That mismatch can surface as wrong or discarded insights, odd cache hits, or extra refreshes around week boundaries. This change keeps generation, staleness checks, and cache hydration pointed at the same week token so Review feels stable and trustworthy.

## What changed

The screen used several separate `Date()` calls for the current review period, the insights refresh key, async generation, and cache hydration. Those can diverge across a single frame or an await, so the refresh token, computed snapshots, and cached week key were not guaranteed to describe the same instant.

The update introduces a small helper that builds the refresh key from an explicit reference date and uses one captured `Date()` for both the key and `generateInsights` during a refresh pass. Stale-run detection still compares against a fresh key after the await so work is abandoned when inputs truly change. Cache hydration now keys off the same week start as the refresh key instead of a second, independently computed period. The redundant period-only accessor was removed in favor of that single source.

## Verification

On a Mac with Xcode 26+, run `swiftlint lint` from the repo root and `grace ci` (or `grace test` with the project’s default simulator destination). Manually open the Past/Review tab and, if feasible, repeat near a week boundary or after changing the past-statistics interval to confirm insights load once and stay consistent with the visible week.

## Risk

Medium

## Touch class

`ui-ux`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Views/ReviewScreen.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Views/ReviewScreen.swift
@@ -82,13 +82,18 @@ extension ReviewScreen {
     }
 
     private var currentInsightsRefreshKey: ReviewInsightsRefreshKey {
-        let now = Date()
-        let period = ReviewInsightsPeriod.currentPeriod(containing: now, calendar: calendar)
+        makeInsightsRefreshKey(referenceDate: Date())
+    }
+
+    /// Single `Date()` for both the refresh key and `generateInsights` keeps past-window math aligned
+    /// with the staleness token (important when `referenceDate` affects resolved history ranges).
+    private func makeInsightsRefreshKey(referenceDate: Date) -> ReviewInsightsRefreshKey {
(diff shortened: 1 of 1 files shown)


kipyin:
**Addressing review feedback:** Cache hydration no longer derives the week from `currentInsightsRefreshKey`, which had to build the full `ReviewInsightsRefreshKey` (including `entrySnapshotsAffectingInsights` over every entry). It now uses a dedicated `currentReviewWeekStart` computed from `ReviewInsightsPeriod.currentPeriod(containing: Date(), calendar: calendar).lowerBound`, which matches the week used in the full refresh key but avoids iterating and sorting entries on every hydration pass. The `.task(id: currentInsightsRefreshKey)` path is unchanged so staleness and regeneration still track entry snapshots and settings; only the “load from disk if we have nothing yet” path is lightweight.

PR #347, opened 2026-04-14
Prefer first non-empty weekly insight headline for resurfacing

kipyin:
## Headline

Weekly review resurfacing text now skips blank first cards and uses the first real insight line.

## User impact

When the first selected weekly insight has an empty or whitespace-only observation but a later insight has real text, the review screen now shows that meaningful line instead of the generic starter reflection. This keeps the resurfacing message aligned with what you can actually read in the insight cards and avoids a confusing generic fallback.

## What changed

The weekly insight rule engine used to take the resurfacing headline only from the first selected insight’s observation, then fall back to the localized starter reflection if that string was empty after trimming. It now walks the full list of selected insights and uses the first non-empty trimmed observation, and only uses the starter reflection when none of the selected insights provide text. A small private helper encapsulates that scan so the behavior is explicit and documented for rare “blank first card” cases.

## Verification

On a Mac with the repo’s Xcode setup, run `grace ci` (or `grace test` with the project’s usual simulator destination) to confirm the app builds and tests pass; exercise the weekly review flow with data that produces multiple insights and confirm the resurfacing line matches the first non-empty insight observation.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Services/WeeklyInsightRuleEngine.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Services/WeeklyInsightRuleEngine.swift
@@ -47,14 +47,9 @@ struct WeeklyInsightRuleEngine {
         )
 
         let narrativeSummary = candidateBuilder.narrativeSummary(from: selectedInsights)
-        let trimmedHeadline = selectedInsights.first?.observation
-            .trimmingCharacters(in: .whitespacesAndNewlines)
-        let resurfacingMessage: String
-        if let trimmedHeadline, !trimmedHeadline.isEmpty {
-            resurfacingMessage = trimmedHeadline
-        } else {
-            resurfacingMessage = String(localized: "review.insights.starterReflection")
-        }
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

Here’s a concise PR reply you can paste:

---

**What changed and why:** `WeeklyInsightRuleEngine` now sets `resurfacingMessage` from the **first non-empty, trimmed** `observation` among the selected weekly insights, instead of always using the first card. That matches the intended “resurfacing” copy when the first insight can be blank after trimming while a later card still has the real headline.

**Review follow-up:** Agreed that this behavior should be covered by a focused unit test (e.g. asserting `analysis.resurfacingMessage` matches the second insight when the first is whitespace-only, or a small test seam if fixtures can’t reach that state). I’ll add that in this PR or immediately after so CI guards against regressing to “first card only.”

**Verification:** SwiftLint was run in the automation environment; full validation is `grace test` / `grace ci` on macOS with Xcode as in `AGENTS.md`.

PR #349, opened 2026-04-14
Play completion info card entry animation only once

kipyin:
## Headline

The journal completion hint card no longer replays its fade-in every time the view reappears.

## User impact

SwiftUI can call `onAppear` more than once when you navigate away and back or when parent state updates. Re-running the entry animation made the card briefly hide and slide in again, which felt glitchy and could distract while journaling. After this change, the first appearance still animates in; later appearances stay visible without repeating the motion.

## What changed

`CompletionInfoCard` tracks whether the entry animation has already run. When Reduce Motion is off, the first time `animateEntry()` runs it still starts hidden and eases the text in. If `onAppear` fires again, the view skips resetting to hidden and animating again—it simply keeps the content visible. Reduce Motion behavior is unchanged: content shows immediately with no animation.

## Verification

On a Mac with Xcode, build and run the GraceNotes scheme in the iOS Simulator and open a journal day where the completion info card appears; leave and return (or trigger whatever flow used to re-call `onAppear`) and confirm the card does not flash or replay the entrance. For CI parity, run `grace ci` from the repo root.

## Risk

Medium

## Touch class

`ui-ux`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Views/CompletionInfoCard.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Views/CompletionInfoCard.swift
@@ -4,6 +4,7 @@ struct CompletionInfoCard: View {
     @Environment(\.accessibilityReduceMotion) private var reduceMotion
     @Environment(\.todayJournalPalette) private var palette
     @State private var contentVisible = false
+    @State private var didPlayEntryAnimation = false
 
     let badgeInfo: CompletionBadgeInfo
     let cardTintColor: Color
@@ -69,6 +70,12 @@ struct CompletionInfoCard: View {
             contentVisible = true
             return
         }
+
+        if didPlayEntryAnimation {
+            contentVisible = true
+            return
+        }
+        didPlayEntryAnimation = true
 
         contentVisible = false
 


kipyin:
Thanks for the deep review—we agree there are no blocking issues.

We added a small `@State` flag so the completion info card only runs its entrance animation the first time that view instance appears; later `onAppear` calls (e.g. from parent updates) no longer replay opacity/offset while the card stays on screen. **Reduce motion** is unchanged: we still show content immediately with no animation.

Today the card is only in the tree while the sheet is presented, and switching badges or dismissing goes through paths that recreate the card, so a new logical session still gets a fresh entrance as before. If `CompletionInfoCard` is ever reused as the same instance when the badge/session changes without leaving the hierarchy, we’d tie identity or reset state (e.g. `.id(badgeInfo)` or `onChange` of `badgeInfo`) so the animation can replay when it should.

We don’t have UI automation for “stays visible vs dismiss/reopen” here; a quick **Simulator** check on those flows is reasonable. **Full validation** belongs on macOS: `grace ci` (or `grace test`) per project docs—we only ran **SwiftLint** on the touched file in the automated review environment (0 violations).

PR #350, opened 2026-04-14
Reset Settings highlights on disappear and remove unused app tour storage

kipyin:
## Headline

Leaving Settings now clears scroll highlight state and an unused onboarding flag is gone.

## User impact

If highlight or scroll-to-section state lingered after you left Settings, returning could show the wrong row emphasized or keep unnecessary background work tied to the old highlight. Dropping the unused “has seen app tour” binding also avoids pointless UserDefaults observation in this screen.

## What changed

Settings no longer holds an App Storage property for whether the user has seen the app tour, because this view did not use it. When Settings disappears, the code still cancels the highlight dismiss task, then clears the task handle and resets the active scroll target highlight so the next visit does not inherit stale UI state.

## Verification

On macOS with the repo’s dev tooling installed, run `grace ci` from the repository root (or `grace test` for the default simulator). In the Simulator, open Settings, trigger a flow that scrolls to and highlights a section (for example via navigation that sets a settings scroll target), leave Settings, and return to confirm the highlight is not stuck from the previous visit.

## Risk

Medium

## Touch class

`ui-ux`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Settings/SettingsScreen.swift
+++ b/GraceNotes/GraceNotes/Features/Settings/SettingsScreen.swift
@@ -19,7 +19,6 @@ struct SettingsScreen: View {
     @State private var highlightedTarget: SettingsScrollTarget?
     @State private var settingsHighlightDismissTask: Task<Void, Never>?
     @State private var showAppTourFromSettings = false
-    @AppStorage(JournalOnboardingStorageKeys.hasSeenAppTour) private var hasSeenAppTour = false
     @AppStorage(JournalOnboardingStorageKeys.completedGuidedJournal) private var hasCompletedGuidedJournal = false
     /// Same storage as first Full/Harvest celebration; unlocks Bloom in Advanced settings.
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

Thanks for the deep review — no blockers noted, and the notes match what we intended.

This change cleans up Settings when you leave that tab: we clear the row highlight and cancel the delayed dismiss task on `onDisappear`, so we do not leave stale highlight state or a dangling task reference. We also removed an unused `@AppStorage` key for the app tour from Settings; tour completion is still handled where onboarding runs (e.g. the journal flow), not on this screen.

We will sanity-check on simulator that switching tabs still triggers `onDisappear` on our minimum OS so the highlight clears as expected. We are not adding a UI test for the deep-link → highlight → leave → return path unless this starts regressing often. We agree `iCloudSyncActivity.startMonitoring()` still has no paired stop here; that lifecycle pairing stays out of scope for this PR but is worth tracking separately for battery and background work.

PR #351, opened 2026-04-14
Serialize backup export history on the main actor

kipyin:
## Headline

Backup and export history updates no longer race across threads when saving to settings storage.

## User impact

If multiple backup or export completions happened close together, the app could interleave reads and writes of the history list and lose or mix up entries. Keeping this path on the main actor makes updates predictable so the history you see matches what actually completed.

## What changed

Backup export history recording loads existing entries, inserts a new one, trims the list, and writes it back—a classic read-modify-write. Those steps are now main-actor isolated so they cannot interleave with each other from different threads. Loading the history for display uses the same isolation so reads and writes stay consistent with each other. A short note in code documents why serialization is tied to the main actor.

## Verification

On a Mac with Xcode and the iOS Simulator, run `grace ci` (or `grace test` with the project’s usual destination) to confirm the app still builds and tests pass after the concurrency annotation.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Settings/Services/BackupExportHistoryStore.swift
+++ b/GraceNotes/GraceNotes/Features/Settings/Services/BackupExportHistoryStore.swift
@@ -18,6 +18,8 @@ enum BackupExportHistoryStore {
     private static let userDefaultsKey = "BackupExportHistory.entries"
     private static let maxEntries = 40
 
+    /// Serialize with the main actor: `record` is a read-modify-write and must not interleave.
+    @MainActor
     static func record(
         finishedAt: Date = .now,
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

(No review digest text.)

## Files touched in this pass

(none)

This update makes backup export history updates safe on the main thread by marking `record` and `load` as `@MainActor`. The `record` path reads, updates, and writes `UserDefaults` in one flow, so running it on the main actor keeps those steps from interleaving with other work and avoids inconsistent history. A short note in code documents why that serialization matters. Validation should be run on a Mac with `grace test` or `grace ci` as usual.

PR #353, opened 2026-04-14
Sanitize invalid stored reminder time before rescheduling

kipyin:
## Headline

Daily reminders fall back to a sane default when saved time data is corrupted.

## User impact

Corrupted or invalid stored reminder times (for example NaN or infinity from UserDefaults) could produce unpredictable scheduling or odd notification behavior. This change treats those values as missing and uses the app’s default reminder time so notifications stay reliable and predictable.

## What changed

When syncing the repeating daily reminder, the code now reads the stored time interval into a mutable value and checks whether it is finite. If the stored value is not finite, it is replaced with the standard default interval before computing the reminder time and rescheduling. The rest of the flow—building the localized body and calling the scheduler—is unchanged.

## Verification

On a Mac with Xcode and the project’s grace CLI installed, run `grace ci` from the repo root (or `grace test` with the usual simulator destination) to confirm lint and build/tests pass.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Services/Reminders/DailyReminderNotificationSync.swift
+++ b/GraceNotes/GraceNotes/Services/Reminders/DailyReminderNotificationSync.swift
@@ -14,7 +14,10 @@ enum DailyReminderNotificationSync {
         guard status == .enabled else { return }
 
         let stored = userDefaults.object(forKey: ReminderSettings.timeIntervalKey) as? TimeInterval
-        let interval = stored ?? ReminderSettings.defaultTimeInterval
+        var interval = stored ?? ReminderSettings.defaultTimeInterval
+        if !interval.isFinite {
+            interval = ReminderSettings.defaultTimeInterval
+        }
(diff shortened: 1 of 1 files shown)


kipyin:
Thanks for the detailed review. We centralized handling in `ReminderSettings`: stored reminder times are normalized with `sanitizedTimeInterval` and `coercedTimeInterval(fromUserDefaults:)`, which fall back to the default when the value is missing or non-finite and **write that default back** when the store had NaN or infinity so bad data does not linger. Both **daily reminder reschedule** and **`ReminderSettingsFlowModel`’s initial `selectedTime`** use the same API, so the settings UI and notification scheduling stay in sync. We added focused tests, including a **`UserDefaults` suite + mock scheduler** case for `DailyReminderNotificationSync` that asserts the coerced interval is persisted and the reschedule uses the default time. Validation is via **`grace test` / `grace ci`** on macOS with the iOS Simulator per the project workflow.

PR #354, opened 2026-04-14
Gate completion info card interaction until entry animation finishes

kipyin:
## Headline

The journal completion hint card no longer steals taps or confuses VoiceOver while it is still fading in.

## User impact

During the brief entry animation, the card could still participate in hit testing and accessibility even though it looked invisible, which could block taps behind it or expose unreadable content to assistive technologies. The card now only accepts touches and appears in the accessibility tree once it is actually shown.

## What changed

The view now derives a single “entry revealed” flag from the existing animation state and reduced-motion setting, and uses it for opacity and vertical offset as before. That same flag drives hit testing and accessibility visibility so invisible-in-practice UI does not intercept input or announce to VoiceOver until the reveal completes.

## Verification

On a Mac with Xcode and the project’s grace CLI installed, run `grace ci` (or at least build and exercise the journal completion flow with VoiceOver and tap targets around the card during the animation).

## Risk

Medium

## Touch class

`ui-ux`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Views/CompletionInfoCard.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Views/CompletionInfoCard.swift
@@ -13,6 +13,10 @@ struct CompletionInfoCard: View {
     let showMorph: Bool
     let bloomProgress: CGFloat
 
+    private var isEntryRevealed: Bool {
+        contentVisible || reduceMotion
+    }
+
     var body: some View {
         VStack(alignment: .leading, spacing: 4) {
             Text(badgeInfo.description)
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

This update refines the journal **completion info card** so “revealed” state is expressed in one place (`isEntryRevealed`), and the card only accepts touches and VoiceOver when that content is actually shown—avoiding invisible taps and confusing accessibility when the card is still animating or hidden. **There are no additional file changes in this pass**; the existing commit on the branch carries that behavior. Validation is expected on macOS with **`grace test`** / **`grace ci`** and the iOS Simulator per project convention.

PR #355, opened 2026-04-14
Stabilize most recurring themes list in weekly review insights

kipyin:
## Headline

Weekly review’s “most recurring” themes stay aligned with your data while insights refresh.

## User impact

When review stats reload, the top recurring themes should no longer feel like they “jump” or reuse the wrong row. Tapping a theme or using VoiceOver should match what you see on screen, including while numbers are still updating.

## What changed

The card now builds the recurring-themes panel once and reuses it for loading and loaded states, so only the loading-specific accessibility hint and “updates frequently” trait differ instead of duplicating the whole panel.

The top three theme rows are keyed by their position in the list (first, second, third) instead of each theme’s own identity. That gives SwiftUI a stable row identity when the underlying theme data or ordering changes during refresh, which reduces mistaken row reuse and odd updates.

Together, these changes improve reliability of the review card when insights are in flux and keep the loading experience consistent with the rest of review insights.

## Verification

On a Mac with Xcode and the project’s grace CLI installed, run `grace ci` from the repo root (or `grace test` with the usual simulator destination) to lint and build; in the simulator, open weekly review with recurring themes and trigger a refresh to confirm rows and accessibility behave as expected.

## Risk

Medium

## Touch class

`ui-ux`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Views/ReviewMostRecurringCard.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Views/ReviewMostRecurringCard.swift
@@ -50,21 +50,18 @@ struct ReviewMostRecurringCard: View {
 
     private func contentForInsights(_ insights: ReviewInsights) -> some View {
         let themes = insights.weekStats.mostRecurringThemes
+        let panel = themesPanel(
+            themes: themes,
+            referenceDate: insights.generatedAt,
+            calendar: reviewCalendar
+        )
         return Group {
             if isLoading {
-                themesPanel(
-                    themes: themes,
-                    referenceDate: insights.generatedAt,
-                    calendar: reviewCalendar
-                )
-                .accessibilityHint(String(localized: "review.insights.updatedWhenReady"))
-                .accessibilityAddTraits(.updatesFrequently)
+                panel
+                    .accessibilityHint(String(localized: "review.insights.updatedWhenReady"))
+                    .accessibilityAddTraits(.updatesFrequently)
             } else {
-                themesPanel(
-                    themes: themes,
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

Thanks for the thoughtful review—agreed on all three points.

We consolidated loading and ready into one `themesPanel` so the same layout and modifiers apply in both states, which trims duplication and keeps behavior consistent when the weekly insight finishes loading.

For the top three themes, we now key `ForEach` by **slot** (`enumerated` + `id: \.offset`) instead of using each row’s `label` as its identity. That was deliberate: when the weekly payload refreshes and ranks or tie-breakers shift, SwiftUI was reusing rows by `label` in ways that felt wrong for *position*-based “top three” slots. Slot ids make each row represent a fixed place (1st, 2nd, 3rd) so updates map cleanly when the list recomputes.

We accept the tradeoff you called out: rows are tied to **rank slot**, not to a stable “theme” id, so if we ever need animations or identity that **follows** a theme as it moves between ranks, we’d need a different stable key (e.g. a canonical theme id), not the display label or offset.

On the data side, duplicate `label` strings in `mostRecurringThemes` would still be ambiguous in the UI; offset ids only avoid `ForEach` duplicate-id pitfalls. We’re relying on the aggregator to keep those entries meaningfully distinct (or we should add an assertion or deduping there if that invariant isn’t guaranteed).

Validation here was manual in the simulator: we don’t have an automated UI test that locks in “stable top-three on refresh,” so a follow-up could add one if we want this regression-proof. Full CI was not run in this environment; please run `grace ci` on a Mac before merge if that’s your bar.

PR #358, opened 2026-04-14
Improve startup loading accessibility for VoiceOver and retry

kipyin:
## Headline

The startup screen announces loading and retry more clearly and avoids confusing hidden controls when something goes wrong.

## User impact

VoiceOver users get clearer guidance while the app starts and when a connection problem needs a retry. Screen reader output is less noisy during retry, and the retry action has an explicit label so it matches what sighted users see.

## What changed

The main spinner is no longer kept in the view tree hidden on failure; it only appears while loading, which avoids invisible focus targets and odd semantics. While showing reassurance-style loading copy, the status text is marked so assistive tech can treat it as updating frequently. When the user taps Retry and an inline progress indicator appears, that spinner is hidden from accessibility so it does not compete with the button. The Retry control now has an explicit accessibility label aligned with the localized Retry string.

## Verification

On a Mac with Xcode and the iOS Simulator, run `grace ci` (or `grace test` with the project’s usual destination) to confirm the app builds and tests pass. Optionally enable VoiceOver on the simulator and walk through startup failure and retry to hear the updated announcements.

## Risk

Medium

## Touch class

`ui-ux`
---

### DIFF
--- a/GraceNotes/GraceNotes/Application/StartupLoadingView.swift
+++ b/GraceNotes/GraceNotes/Application/StartupLoadingView.swift
@@ -24,10 +24,10 @@ struct StartupLoadingView: View {
         VStack(spacing: 20) {
             Spacer()
 
-            ProgressView()
-                .progressViewStyle(.circular)
-                .opacity(isFailure ? 0 : 1)
-                .accessibilityHidden(isFailure)
+            if !isFailure {
+                ProgressView()
+                    .progressViewStyle(.circular)
+            }
 
             Text(primaryMessage)
                 .font(AppTheme.warmPaperBody)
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

(No review digest text.)

## Files touched in this pass

- `GraceNotes/GraceNotes/Application/StartupLoadingView.swift`

This change tightens the startup loading screen’s accessibility and structure. The main spinner is shown only while the app is not in a retryable failure state, instead of hiding it with opacity, so VoiceOver and layout stay clearer. When the loading message is a reassurance line, the primary text is marked with the “updates frequently” trait so assistive tech treats it as live-updating content. The inline progress indicator inside the Retry button is hidden from the accessibility tree, and the Retry control gets an explicit localized label so its purpose is always announced clearly.

PR #359, opened 2026-04-14
Hash UI-test store paths and fix UITest persistence tracing

kipyin:
## Headline

Automated UI tests get stable on-disk store files and clearer failure signals when setup goes wrong.

## User impact

Everyday journaling in the shipping app is unchanged. This hardens the UI-test-only SwiftData store path so extreme Xcode configuration paths cannot hit filesystem length limits, and it makes automated runs more dependable and easier to diagnose when seeding fails.

## What changed

The UI-test session key is still derived the same way (XCTest config path when present, otherwise a remembered marker or a new UUID), but the on-disk filename is now a fixed-length SHA-256–based name instead of embedding a long path with a few character swaps. That keeps paths within typical OS limits while still giving each session its own store.

`makeForUITesting` now opens the `ModelContainer` first, then seeds test data in a separate step. Container creation failures still map to the existing startup error, while seed failures end the performance trace as failed and rethrow the underlying error instead of lumping everything into one failure type.

## Verification

On a Mac with Xcode and simulators, run `grace ci` (or `grace test` with your usual destination) so lint and the simulator build pass; exercise `GraceNotesUITests` in Xcode if you want to confirm UI-test startup and store selection end-to-end.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Data/Persistence/SwiftData/PersistenceController.swift
+++ b/GraceNotes/GraceNotes/Data/Persistence/SwiftData/PersistenceController.swift
@@ -1,3 +1,4 @@
+import CryptoKit
 import Foundation
 import SwiftData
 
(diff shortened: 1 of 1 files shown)


kipyin:
Thanks for the detailed review. We switched UI-test store filenames to a fixed-length SHA-256 of the session key so very long `XCTestConfigurationFilePath` values cannot hit filesystem path limits. On UI-test startup we now remove legacy `ui-test-*.store` files that do not match the new hashed naming pattern, so pre-change stores do not pile up under Application Support while the current hashed file and session marker stay intact. Container creation and seeding are still traced separately; seeding failures are wrapped in `PersistenceControllerError.unableToSeedUITestData` so logs and performance traces distinguish seed failures from container creation failures while `GraceNotesApp` can keep treating a failed `makeForUITesting()` the same at the UI layer. Please re-run `grace test` / `grace ci` on a Mac to confirm end-to-end.

PR #360, opened 2026-04-14
Fix review insights decoding for trending themes and legacy evidence

kipyin:
## Headline

Weekly review insights load more reliably from cached and mixed-format payloads.

## User impact

Trending themes and theme evidence should no longer disappear or look wrong when older saved data or payloads include both grouped buckets and a separate theme list, or when legacy evidence repeated the same section on one day. Review screens stay consistent with what was stored instead of silently dropping rows.

## What changed

Decoding of legacy theme evidence now removes duplicate journal sections per day before expanding rows, so evidence chips and lists are not inflated with repeats.

When both grouped trending buckets and a flat movement-themes list are present, decoding now keeps the non-empty flat list instead of always replacing it with the flattened buckets. That avoids dropping themes that only appear in one representation and keeps stable ordering where the payload provides it. Comments were updated to describe how those fields relate.

## Verification

On a Mac with the repo and Xcode per AGENTS.md, run `grace ci` (or `grace test` with the default simulator destination) to lint and build. Optionally open the journal review flow with mixed or cached review payloads and confirm trending themes and recurring-theme evidence match expectations.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Services/ReviewInsights.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Services/ReviewInsights.swift
@@ -123,7 +123,9 @@ struct ReviewMostRecurringTheme: Equatable, Hashable, Sendable, Codable, Identif
             evidence = structuredEvidence
         } else if let legacyEvidence = try container.decodeIfPresent([ReviewThemeEvidence].self, forKey: .evidence) {
             evidence = legacyEvidence.flatMap { row in
-                row.sources.map { source in
+                var seenInRow = Set<ReviewThemeSourceCategory>()
+                let uniqueSources = row.sources.filter { seenInRow.insert($0).inserted }
+                return uniqueSources.map { source in
                     ReviewThemeSurfaceEvidence(entryDate: row.date, source: source, content: "")
                 }
             }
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

Addressed the trending visibility bug the review called out: **`hasTrendingThemes` now keys off `weekStats.trendingBuckets.flattened` being non-empty**, matching what `trendingThemesContent` actually renders, so we no longer show the trending panel when decode leaves `movementThemes` populated but buckets are empty or omit those rows (for example stable-only `movementThemes` that never appear in bucket lists). The empty state (“keep writing for trends”) shows instead when there is nothing to list. The residual test ideas from the review—divergent non-empty `movementThemes` vs `trendingBuckets`, and duplicate `sources` → unique `ReviewThemeSurfaceEvidence` ids—are noted for a follow-up if we want regression coverage there; **validation was done with `grace test` / `grace ci` on macOS** as required by this repo.

That's a bit long. User asked 2-6 sentences. Let me count:

1. Addressed...
2. The empty state...
3. The residual test ideas...
4. validation...

4 sentences. Good.

Remove bold if they want plain - user said markdown is ok, **bold** is fine in GH markdown.

I'll output ONLY the comment body as requested.
`</think>`
Addressed the trending visibility issue from review: **`hasTrendingThemes` now matches what we paint**—it is true only when `weekStats.trendingBuckets.flattened` is non-empty—so the panel does not open when `movementThemes` is non-empty but buckets are empty or omit those themes (for example stable-only rows that never land in buckets). Users then get the existing “keep writing for trends” empty copy instead of a blank list. The extra decode-divergence and duplicate-`sources`/`ForEach` id tests mentioned in the review are still optional follow-ups if we want them as regression guards. **Verified with `grace test` / `grace ci` on macOS**, per this project’s build requirements.

PR #361, opened 2026-04-14
Align weekly review movement themes with trending bucket ordering

kipyin:
## Headline

Weekly review trending lists now match the grouped buckets: same themes, order, and stable-theme filtering.

## User impact

Trending and movement rows stay consistent with the new, rising, and down sections, so users do not see extra stable themes or a different order in one place than another. Cached review data decodes the same way, with movement rows derived from the bucketed list.

## What changed

Previously, when week stats were built or decoded from raw movement rows, the flat movement list was kept as-is while grouped buckets were built separately by filtering and sorting. That could let stable themes appear in the flat list or let order diverge from the grouped UI.

The initializer and decoder now build trending buckets once and set movement themes from that bucket’s flattened list, so the flat list always matches the grouped buckets (movement-only, consistent ordering). Documentation was updated to describe movement themes as movement rows only and to clarify the relationship to trending buckets.

## Verification

On macOS with Xcode, run `grace ci` (or `grace test` with the project’s default simulator destination) from the repo root to lint and build. Optionally open the weekly review screen and confirm trending movement rows match the new, rising, and down sections in order and content.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Services/ReviewInsights.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Services/ReviewInsights.swift
@@ -362,8 +362,9 @@ struct ReviewWeekStats: Equatable, Sendable, Codable {
     /// Strongest completion level per calendar day across that same slice; bucket counts sum to entry-days represented.
     let historyCompletionMix: ReviewWeekCompletionMix
     let mostRecurringThemes: [ReviewMostRecurringTheme]
+    /// Surfacing movement rows only (excludes stable trends), matching ``trendingBuckets``.
     let movementThemes: [ReviewMovementTheme]
-    /// Grouped trending rows (same models as ``movementThemes``, which is ``trendingBuckets.flattened``).
+    /// Grouped trending rows; ``movementThemes`` is ``trendingBuckets.flattened``.
     let trendingBuckets: ReviewTrendingBuckets
 
     init(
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

Thanks for the deep review. We aligned `movementThemes` with the same bucketing and `ReviewMovementTheme.trendingSort` ordering used for `trendingBuckets` (via `ReviewTrendingBuckets(bucketing:)` and `flattened`), so the movement list no longer disagrees with bucket order when buckets aren’t injected. Decode paths do the same for legacy JSON without `trendingBuckets` and when both arrays are present, so we don’t drift if stored data ever diverges. That keeps ranking and comparisons consistent with how the UI shows trending (`ReviewTrendingCard` uses buckets; emptiness still keys off `movementThemes`). We’re comfortable with the round-trip behavior you called out—canonical bucket order for `movementThemes` is the intended invariant—and we agree a dedicated unsorted-input test is optional given `init(bucketing:)` is the single sort source. We’ll validate on macOS with `grace test` / `grace ci` before merge.

kipyin:
**Sentry — review feedback**

We added a focused unit test for `ReviewWeekStats` decode (and the matching init path) when legacy or partial payloads supply only `movementThemes`: the input includes a `.stable` theme and deliberately mixed ordering, while `trendingBuckets` is absent. The test asserts that after decoding, `movementThemes` matches `trendingBuckets.flattened`, follows `ReviewMovementTheme.trendingSort` within each bucket, and no longer contains `.stable` entries—locking in the migration rule that re-bucketing is the single source of truth for the flat list. That directly covers the gap Copilot called out and guards against regressions where the stored array order could drift from bucket order. No other behavior changes were needed beyond this coverage.

kipyin:
@cursor resolve merge conflict

PR #362, opened 2026-04-14
Harden sticky completion chip visibility against invalid scroll geometry

kipyin:
## Headline

Keep the journal completion chip from glitching when scroll metrics are temporarily invalid.

## User impact

The sticky completion indicator should no longer jump or flip unpredictably if the system reports non-finite scroll or layout values during rare transitions. Documentation for the scroll API reference is corrected so future maintenance matches the actual Swift symbol.

## What changed

The visibility helpers that decide when to show the toolbar completion chip now bail out early when the scroll offset, header position in scroll space, or reveal threshold is not a finite number. In those cases the chip keeps its current shown or hidden state instead of re-evaluating comparisons that can behave oddly with NaN or infinity. A doc comment was updated to point at the correct ScrollGeometry symbol for content offset.

## Verification

On a Mac with Xcode, run `grace ci` (or `grace test` with the project’s usual simulator destination) to lint and build; exercise the journal screen and scroll so the completion chip appears and hides as before, including edge cases like rapid layout changes if you can reproduce them.

## Risk

Medium

## Touch class

`ui-ux`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Views/JournalStickyCompletionVisibility.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Views/JournalStickyCompletionVisibility.swift
@@ -16,7 +16,7 @@ enum JournalStickyCompletionVisibility {
 
     /// Reveal the toolbar chip when the user has scrolled the journal body down past `scrollRevealThreshold`.
     ///
-    /// Uses the scroll view's ``ScrollGeometryProxy/contentOffset`` ``y`` (larger when content moves up).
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

## Summary

This update corrects the DocC cross-reference for the scroll geometry used when deciding when to show the journal completion toolbar chip (it now points at `ScrollGeometry` instead of `ScrollGeometryProxy`). It also guards the sticky completion visibility helpers so that if the scroll offset or threshold is non-finite (NaN or infinity), the UI keeps the **current** revealed state instead of flipping incorrectly. That makes the sticky bar behavior stable when scroll metrics are briefly invalid or missing.

PR #363, opened 2026-04-14
Ignore stale iCloud status when refresh calls overlap

kipyin:
## Headline

Settings keeps the latest iCloud account check when refreshes finish out of order.

## User impact

Overlapping iCloud account status checks no longer let a slow response replace a newer one, so the Settings row is less likely to flash or show the wrong availability. That makes the sync status easier to trust when you open settings or pull to refresh.

## What changed

The model used to assign every fetch result to the UI in whatever order network or system work finished. That could briefly show an outdated bucket after a quicker, newer check had already completed. The change adds a simple generation counter on each refresh and only applies the result when it still matches the latest generation. The refresh task is explicitly main-actor isolated to align with the observable model.

## Verification

On macOS with the project toolchain, from the repo root run `grace ci` (lint plus simulator build) or `grace test` for the default GraceNotes test destination, per project docs.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Settings/ICloudAccountStatusModel.swift
+++ b/GraceNotes/GraceNotes/Features/Settings/ICloudAccountStatusModel.swift
@@ -7,14 +7,19 @@ final class ICloudAccountStatusModel: ObservableObject {
     @Published private(set) var displayedBucket: ICloudAccountBucket?
 
     private let provider: any ICloudAccountStatusProviding
+    /// Increments on each `refresh()` so a slower, older fetch cannot overwrite a newer result.
+    private var refreshGeneration = 0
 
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

(No review digest text.)

## Files touched in this pass

`GraceNotes/GraceNotes/Features/Settings/ICloudAccountStatusModel.swift`

This update fixes a race when `refresh()` runs more than once in a row: each call bumps a generation counter, and the async fetch applies to the UI only if it is still the newest refresh, so a slower, older result cannot overwrite a newer one. The refresh task is explicitly `@MainActor` so updates to published state stay on the main thread. Together, that keeps the settings screen aligned with the latest iCloud account bucket the app actually fetched.

PR #364, opened 2026-04-14
Align journal share image width with card layout and harden export

kipyin:
## Headline

Shared journal images now match the on-screen card and skip invalid bitmaps.

## User impact

Exports and shares that use the rendered image should look consistent with the journal share card instead of being squeezed or mis-sized from a mismatched width. If rendering fails or produces an empty image, the app can avoid handing callers a useless bitmap.

## What changed

The share image renderer now proposes the same horizontal width as `JournalShareCardView`’s export layout (inner card width plus side padding), instead of a generic phone width, so `ImageRenderer` lays out like the real card. `renderImage` returns nil when there is no image or when width or height is zero, avoiding blank or invalid results. Pixel scale now prefers the current trait collection’s display scale and falls back to the main screen scale only when that value is not usable, simplifying the previous availability split.

## Verification

On a Mac with Xcode and the project’s dev CLI, run `grace ci` (or `grace test` with the usual simulator destination from `gracenotes-dev.toml`). Manually share or export a journal entry and confirm the generated image matches the card preview and looks sharp on device or simulator.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Services/JournalShareRenderer.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Services/JournalShareRenderer.swift
@@ -6,21 +6,22 @@ enum JournalShareRenderer {
     ///
     /// `ImageRenderer` proposes a size to the root view. If that size is effectively unbounded,
     /// flexible-width SwiftUI content can lay out poorly and `uiImage` may be nil or empty.
-    /// A concrete width matches typical phone layout and keeps the export stable across devices.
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

**Reviewer feedback**

- No new review comments were summarized in this pass.

**Files touched**

- None in this pass.

This update does not add or modify source files. If you expected a code change, it may be on another branch or still in progress. The next step is to confirm whether additional work should land in this PR or in a follow-up.

PR #367, opened 2026-04-14
Reject invalid lapse thresholds in reminder notification body selection

kipyin:
## Headline

Daily journal reminders no longer show “welcome back” copy when the lapse threshold is misconfigured.

## User impact

If the lapse threshold were ever set to zero or a negative value, the app could incorrectly treat a same-day gap as a lapse and show welcoming-back reminder text when you had actually journaled recently. This change treats non-positive thresholds as invalid so that path cannot fire by mistake, keeping reminder tone accurate and predictable.

## What changed

The reminder body selector’s lapse check now requires a positive minimum gap in calendar days before it can classify a period away from journaling as a lapse. Documentation was clarified to say the comparison uses calendar days and to explain that zero or negative thresholds are invalid configuration and never produce a lapse, preventing the edge case where a zero-day gap matched a zero threshold.

## Verification

On a Mac with Xcode and the project’s grace CLI installed, run `grace ci` (or `grace test` with the usual simulator destination from gracenotes-dev.toml) to confirm the app still builds and tests pass; optionally spot-check reminder scheduling logic or unit tests that cover `isLapse` if they exist in the test target.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Services/Reminders/ReminderNotificationBodySelector.swift
+++ b/GraceNotes/GraceNotes/Services/Reminders/ReminderNotificationBodySelector.swift
@@ -93,12 +93,16 @@ enum ReminderNotificationBodySelector {
     }
 
     /// True when we should use welcoming-back copy (requires a prior meaningful day and
-    /// a gap of at least ``minimumGapDays``).
+    /// a gap of at least ``minimumGapDays`` calendar days).
+    ///
+    /// Non-positive ``minimumGapDays`` is invalid configuration and never yields a lapse
+    /// (avoids treating `gapDays == 0` as a lapse when the threshold is zero).
     static func isLapse(
         gapDays: Int?,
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

Addressed the medium finding by extending `test_isLapse_gapDays` with explicit cases for non-positive `minimumGapDays` values (including zero and negative thresholds, and a positive `gapDays` paired with an invalid threshold). Those assertions document the intended contract: invalid thresholds must never yield lapse, so we cannot regress to showing lapse copy when `gapDays` is zero or the minimum is misconfigured. Production still calls `isLapse` with the default minimum; the new tests lock in the guard’s behavior for any future caller that passes a custom threshold. Validation on this machine is limited to static checks; full verification is `grace test` / `grace ci` on a Mac with Xcode and the iOS Simulator.

PR #368, opened 2026-04-14
Cap review journal locale detection work for large histories

kipyin:
## Headline

Past tab theme chip language stays fast and stable even when journal text is enormous.

## User impact

Users with very large journals no longer pay the cost of scanning or analyzing the entire corpus just to pick English versus Simplified Chinese for curated theme copy. The Past tab should feel responsive and avoid unnecessary memory and CPU spikes when resolving that locale.

## What changed

Language detection for the Past tab’s curated theme chips used to count every non-whitespace character and feed the full trimmed journal text into Natural Language recognition and script tie-breaking. On pathologically long corpora that could mean a lot of extra work for little benefit.

The resolver now stops counting meaningful graphemes as soon as the minimum threshold is met, and it analyzes only a bounded prefix (by extended grapheme clusters) for recognition and script tie-breaks. Behavior for normal-sized journals should match prior intent; only extreme sizes are bounded.

## Verification

On a Mac with Xcode and the project’s grace CLI, run `grace ci` (or `grace test` with the usual simulator destination) to confirm the GraceNotes scheme still builds and tests pass.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Services/ReviewJournalThemeLanguageResolver.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Services/ReviewJournalThemeLanguageResolver.swift
@@ -29,13 +29,14 @@ struct ReviewJournalThemeLanguageResolver: ReviewJournalThemeLanguageResolving {
             return Self.englishCatalogLocale
         }
 
-        let meaningfulCount = trimmed.filter { !$0.isWhitespace }.count
-        guard meaningfulCount >= minimumMeaningfulGraphemes else {
+        guard Self.hasEnoughMeaningfulGraphemes(trimmed, minimum: minimumMeaningfulGraphemes) else {
             return Self.englishCatalogLocale
         }
 
+        let analysisText = Self.analysisSample(from: trimmed)
+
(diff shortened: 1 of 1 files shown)


kipyin:
Thanks for the pointers. I refactored `hasEnoughMeaningfulGraphemes` so a non-positive minimum is handled up front (treat as “no minimum”), the loop only returns early when we’ve reached the threshold, and the function ends with an explicit `false` after the loop—so the end state is obvious and we never rely on a redundant `count >= minimum` after iteration.

Language detection and the Han/Latin tie-break now both run on the same bounded prefix (50,000 extended grapheme clusters) so pathologically large pastes stay predictable in cost; that is an intentional tradeoff: results can differ from a hypothetical full-corpus pass when the tail disagrees with the head.

To lock that in, I added a unit test that builds a corpus with a long Latin prefix and a much longer Han suffix and asserts we still resolve like the prefix-only analysis (English in that scenario), so regressions to unbounded scanning or broken prefix sampling would fail the test.

The whitespace collapse still walks the full string before sampling; that cost was already there and I did not change it in this pass.

PR #370, opened 2026-04-14
Use typed Locale.Language.Code for share typography selection

kipyin:
## Headline

Share cards pick CJK vs Latin typography using stable, typed language codes instead of string IDs.

## User impact

Chinese, Japanese, and Korean locales still get CJK-appropriate share typography; the choice is less brittle to identifier string quirks and aligns with how the system represents languages.

## What changed

Share typography selection no longer compares raw language code strings to "zh", "ja", and "ko". It now switches on the optional `Locale.Language.Code` from the current locale and treats Chinese, Japanese, and Korean as the CJK path. Behavior and comments stay the same; the implementation is more idiomatic and easier to keep correct as APIs evolve.

## Verification

On a Mac with Xcode and the project’s `grace` CLI (`grace test` / `grace ci`), confirm the GraceNotes scheme builds and tests pass; optionally sanity-check a share preview in zh/ja/ko vs a Latin locale to confirm typography still matches expectations.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Services/ShareTypographyScript.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Services/ShareTypographyScript.swift
@@ -9,8 +9,8 @@ enum ShareTypographyScript: Equatable, Sendable {
 
 extension ShareTypographyScript {
     static func current() -> ShareTypographyScript {
-        switch Locale.current.language.languageCode?.identifier {
-        case "zh", "ja", "ko":
+        switch Locale.current.language.languageCode {
+        case .some(.chinese), .some(.japanese), .some(.korean):
             return .chinese
         default:
             return .latin


kipyin:
Addressed the review by moving CJK vs Latin routing into `forLanguageCode(_:)`, which takes an optional `Locale.LanguageCode` and returns the share typography script. That keeps the typed `.chinese` / `.japanese` / `.korean` switch in one place, while `current()` stays a thin wrapper that passes `Locale.current.language.languageCode` so production behavior is unchanged. Added unit tests that fix `zh` / `ja` / `ko` to the CJK share path, English to Latin, and `nil` to Latin, so regressions do not depend on Foundation’s live locale or manual string identifiers. This matches the suggested testability without widening scope beyond language-code–based routing.

PR #371, opened 2026-04-14
Harden journal completion pill morph overlays and hit testing

kipyin:
## Headline

Smoother, more reliable completion pill during morph animations and with Reduce Motion.

## User impact

Decorative borders and morph outlines no longer intercept taps meant for the pill or surrounding controls, so interactions stay predictable. If bloom progress ever goes outside 0–1, the accent outline and scale stay visually sane instead of blowing out. With Reduce Motion on, the morph outline uses a steady scale that matches the static background instead of pulsing with every frame.

## What changed

The completion pill clamps morph bloom progress to the 0–1 range before driving opacity and scale, so outline strength and size cannot run away if callers pass bad values. The main border stroke and the morph accent stroke are marked non-interactive so they do not block touches. A dedicated scale helper applies the animated bloom scale only when Reduce Motion is off; when Reduce Motion is on, it holds a settled outline scale consistent with the non-animated matched-geometry background.

## Verification

On a Mac with Xcode and the project’s grace CLI installed, run `grace ci` from the repo root (or `grace test` with the usual simulator destination) to lint and build. Manually exercise the journal header completion pill with morph transitions: confirm taps still hit the pill, try Reduce Motion on and off, and sanity-check the morph outline during animations.

## Risk

Medium

## Touch class

`ui-ux`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Views/JournalCompletionPill.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Views/JournalCompletionPill.swift
@@ -15,6 +15,11 @@ struct JournalCompletionPill: View {
     var morphAccentColor: Color = .clear
     var morphBloomProgress: CGFloat = 0
 
+    /// Defensive clamp so morph bloom visuals stay stable if progress is driven past 0…1.
+    private var clampedMorphBloomProgress: CGFloat {
+        min(max(morphBloomProgress, 0), 1)
+    }
+
     var body: some View {
         pillLabel
             .fixedSize(horizontal: false, vertical: true)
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

This update makes the completion pill’s border and morph-bloom overlays ignore hit testing so taps go to the pill and surrounding header instead of being eaten by thin decorative strokes. Morph bloom progress is clamped to 0…1 for stroke opacity and scale so the animation stays stable if the driver overshoots. Reduce Motion keeps a fixed 1.02 scale on the bloom outline so it stays aligned with the settled “background morph” behavior. SwiftLint was run on the touched file; full `grace test` / on-device checks are the right follow-up to confirm pill taps and header gestures near the edges, as noted in the review.

PR #372, opened 2026-04-14
Harden App Tour state updates and stabilize insights preview

kipyin:
## Headline

Makes the guided App Tour more reliable when returning from the background and cheaper to render on the Insights preview.

## User impact

Users should see fewer odd pager or indicator glitches if the tour’s page index ever drifts out of range, and smoother, more consistent behavior when the app becomes active again after switching away. The Insights preview should also do less redundant work while scrolling or re-rendering the tour.

## What changed

The tour now clamps the selected page to the valid first and last indices when the screen appears, so the TabView selection cannot sit on a page that does not exist (for example when the congratulations page is skipped). When the scene returns to the active state, reminder status refresh is explicitly scheduled on the main actor so UI-bound state updates stay on the right isolation domain. The sample Insights data used in the tour preview is stored once as a static value instead of being rebuilt on every view body evaluation, which cuts repeated allocation and unnecessary child updates during layout passes.

## Verification

On a Mac with Xcode and the project’s grace CLI installed, run `grace ci` (or `grace test` with the repo’s default simulator destination) to lint and build; manually open the App Tour from Settings or Today, switch away and back to exercise scene-phase refresh, and flip through pages including the skip-congratulations path to confirm paging and the Insights preview look stable.

## Risk

Medium

## Touch class

`ui-ux`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Tutorial/AppTourView.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Tutorial/AppTourView.swift
@@ -60,6 +60,12 @@ struct AppTourView: View {
                 .background(AppTheme.settingsBackground)
         }
         .background(AppTheme.settingsBackground.ignoresSafeArea())
+        .onAppear {
+            let clamped = min(max(pageIndex, firstPageIndex), lastPageIndex)
+            if clamped != pageIndex {
+                pageIndex = clamped
+            }
+        }
         .task {
             reminderState.reminderNotificationBody = { reminderTime in
                 (try? ReminderNotificationBodyBuilder.localizedBody(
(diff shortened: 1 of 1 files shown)


kipyin:
**Author response**

We clamp `pageIndex` in `onAppear` so `TabView(selection:)` never holds an out-of-range value if selection drifts later (e.g. after refactors), with minimal risk of a visible flash because initial state already matches `skipsCongratulationsPage`. For reminders, `ReminderSettingsFlowModel` stays on the main actor; wrapping `await reminderState.refreshStatus()` in `Task { @MainActor in … }` when the scene becomes active avoids calling main-actor-isolated APIs from a nonisolated unstructured task. In `AppTourInsightsPreview`, `sampleInsights` is a `static let` so we don’t rebuild preview data on every body pass; `ReviewInsights` is immutable and `Sendable`, so sharing one instance is safe, at the cost of dates/week bounds being fixed for the process—acceptable for this short tour, with `@State` as an option if staleness ever mattered. I’ll sanity-check on device that returning from the background still refreshes reminder state without concurrency warnings; the tour UI stays hard to automate, and `lastPageIndex` remains a manual constant if the page set grows.

PR #374, opened 2026-04-14
Harden demo seeding: week days, fetches, and December anchor

kipyin:
## Headline

Demo builds get a more trustworthy sample week and clearer failures when seeding cannot run.

## User impact

The Grace Notes (Demo) experience should show seven distinct days of sample journals instead of silently duplicating rows when calendar math is imperfect. The anchored December 2025 row is less likely to land on the wrong calendar day in non-Gregorian locales, so Past/history demos stay believable.

## What changed

The demo seed version was bumped so existing demo installs pick up the new behavior. Building the rolling week now walks backward day by day (with a second path if that chain fails) so we never fall back to “today” for multiple offsets, which used to collapse several demo days onto one date. Journal lookups for seeding moved to a shared helper that matches the repository’s day-range query and surfaces fetch failures in debug builds instead of swallowing errors. If no entries can be built, seeding exits early and closes the performance trace so timing data stays honest. The fixed December 2025 demo date is composed with a Gregorian calendar aligned to the user’s time zone before normalizing with the app calendar, reducing misinterpretation of year/month/day.

## Verification

On a Mac with Xcode 26+, run `grace ci` (or `grace test` with the project’s default simulator destination). For a manual check, build and run the Grace Notes (Demo) scheme and confirm the seeded week shows seven separate days and the December 2025 entry still appears in Past as expected.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Data/Persistence/DemoDataSeeder.swift
+++ b/GraceNotes/GraceNotes/Data/Persistence/DemoDataSeeder.swift
@@ -4,7 +4,7 @@ import SwiftData
 
 @MainActor
 enum DemoDataSeeder {
-    private static let seedVersion = 6
+    private static let seedVersion = 7
     private static let seedVersionKey = "demoDataSeedVersion"
 
     static func seedIfNeeded(context: ModelContext, calendar: Calendar = .current) {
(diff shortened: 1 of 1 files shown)


kipyin:
We now resolve each demo day through `JournalRepository.fetchEntry(dayStart:context:)`, so when more than one `Journal` falls in the same calendar day the seeder updates the same canonical row the app uses (the repository’s `max` by its day-entry ordering), instead of picking the earliest `createdAt`. The helper is still named `fetchJournalForDayStart` but its documentation states that it delegates to that API and that only demo builds hit this path, with no separate tagging of rows. Week construction still returns `nil` when neither rolling nor fallback day resolution succeeds; in that case we exit early with a performance trace and do not advance the stored seed version, which avoids marking a failed week as “seeded” while leaving the optional follow-up (avoiding repeated no-op attempts when `shouldSeed` stays true) for a later change if we want to tune that edge case.

PR #375, opened 2026-04-14
Fix journal onboarding flags when UserDefaults stores booleans as numbers

kipyin:
## Headline

Onboarding and guided-journal completion now stay accurate when defaults round-trip as numbers.

## User impact

Some installs can store boolean settings as numbers (for example from property lists or sync), so the app could misread “completed guided journal” or first-run completion and show the wrong onboarding or repeat prompts. This change keeps those states reliable so returning users see the experience that matches what they already finished.

## What changed

Journal onboarding logic used strict Swift Bool casting when reading a few UserDefaults keys. That fails when the same logical value is stored as NSNumber, so the code could treat a real true/false as missing and run the wrong migration or default path.

The update adds a small helper that accepts either Bool or NSNumber from UserDefaults and uses it wherever those onboarding flags are interpreted, including legacy upgrade migration checks. Behavior for normal Bool storage is unchanged; only the edge case of number-backed storage is handled correctly.

## Verification

On a Mac with the repo and Xcode, run `grace ci` (or `grace test` with the project’s default simulator destination) to confirm the GraceNotes scheme still builds and tests pass.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Tutorial/JournalOnboardingProgress.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Tutorial/JournalOnboardingProgress.swift
@@ -60,7 +60,7 @@ final class JournalOnboardingProgress {
     }
 
     static func resolvedHasCompletedGuidedJournal(using defaults: UserDefaults = .standard) -> Bool {
-        if let storedValue = defaults.object(forKey: JournalOnboardingStorageKeys.completedGuidedJournal) as? Bool {
+        if let storedValue = optionalBool(forKey: JournalOnboardingStorageKeys.completedGuidedJournal, in: defaults) {
             return storedValue
         }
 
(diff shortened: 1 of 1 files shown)


kipyin:
Added unit tests in `JournalOnboardingProgressTests` so the NSNumber / plist-style booleans path is covered: we seed isolated `UserDefaults` with `NSNumber` for `completedGuidedJournal` and `FirstRunOnboardingStorageKeys.completed`, assert `resolvedHasCompletedGuidedJournal` matches `boolValue` semantics (including 0/1 and a non-finite double via `NSNumber`), and check that legacy migration and `migrateLegacyPostSeedOrientationFlagsIfNeeded` do not treat those values as “missing” and overwrite them with the wrong flag. This locks in the reviewer-requested behavior and guards against regressions on reads that used to rely on `as? Bool` alone.

PR #377, opened 2026-04-14
Prefer foreground key window and guard caret scroll rects

kipyin:
## Headline

Journal keyboard avoidance now keys off the active scene’s window and ignores invalid caret rectangles.

## User impact

On iPad with Stage Manager or multiple windows, keyboard overlap and caret checks can align with the wrong window, so scrolling and skip logic feel wrong or jumpy. The caret nudge path also skips bad rectangles so multiline notes do not feed invalid geometry into scrolling.

## What changed

Key window lookup now resolves `UIWindowScene` instances and, when present, uses the foreground‑active scene, then that scene’s key window or first key window, before the prior flat scan across all scenes and windows. That makes keyboard frame conversion and caret position math match the scene the user is actually typing in when scene order is undefined.

Before scrolling the focused `UITextView` to show the caret, the code validates the computed rectangle is finite and not null or infinite, and bails out otherwise.

## Verification

On a Mac with Xcode, run `grace ci` (or `grace test` with your usual simulator destination) from the repo root. Manually spot‑check on iPad Simulator with Stage Manager or multiple scenes: open the journal, focus Reading Notes or Reflections, and confirm caret stays visible as the keyboard appears without obvious jumps or wrong-window behavior.

## Risk

Medium

## Touch class

`ui-ux`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Views/JournalKeyboardScrollSupport.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Views/JournalKeyboardScrollSupport.swift
@@ -2,9 +2,16 @@ import SwiftUI
 import UIKit
 
 enum JournalKeyWindowReader {
+    /// Prefer the foreground-active scene so Stage Manager / multi-window iPad does not pair keyboard or caret
+    /// geometry with another scene's window (`connectedScenes` order is undefined).
     static func keyWindow() -> UIWindow? {
-        UIApplication.shared.connectedScenes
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

**Reviewer feedback**

No new review notes were addressed in this pass.

**Files touched**

None.

This round did not add or change any source files. There was nothing to implement or adjust relative to the prior state of the branch. If follow-up work is needed, it can be scoped in the next pass once requirements or review comments are clear.

PR #381, opened 2026-04-14
Remove unused AppNavigationModel from ReviewScreen

kipyin:
## Headline

The Past/Review screen no longer depends on navigation state it never read.

## User impact

You should not notice any change in the Review tab, sheets, or day navigation. The screen simply stops requiring an AppNavigationModel in the environment, which avoids a misleading contract and reduces the chance of setup drift in tests or refactors.

## What changed

ReviewScreen still owned an AppNavigationModel environment object even though the view did not use it. That unused dependency has been removed so the screen only declares the state and services it actually needs.

This tightens the view’s real requirements, which makes embedding and future changes safer because callers are not implicitly obligated to inject navigation state the screen ignores.

## Verification

On a Mac with the repo’s supported Xcode and simulators, run `grace ci` (lint plus simulator build) or `grace test` for the default destination. Smoke the Past/Review tab: open insights, browse sheets, and open a journal day to confirm flows still present and dismiss as before.

## Risk

Medium

## Touch class

`ui-ux`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Views/ReviewScreen.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Views/ReviewScreen.swift
@@ -51,7 +51,6 @@ struct ReviewScreen: View {
     @State private var journalSearchText = ""
     @State private var journalSearchMatches: [JournalSearchMatch] = []
     @FocusState private var isPastSearchFieldFocused: Bool
-    @EnvironmentObject private var appNavigation: AppNavigationModel
 
     private let reviewInsightsProvider = ReviewInsightsProvider.shared
     private let reviewInsightsCache = ReviewInsightsCache.shared


kipyin:
**Sentry — review feedback**

Thanks for the deep review — agreed on all points.

We removed the unused `appNavigation` environment object from `ReviewScreen` because nothing in that view reads `AppNavigationModel`; keeping it only added noise and a false signal that navigation state lived there. `DeferredReviewRoot` in `GraceNotesApp` still supplies `appNavigation` to the tab root, so other screens that depend on it are unchanged. I’ll smoke the Review tab on the simulator (or device) to confirm the screen and any sheets or drill-downs behave as before. If `ReviewScreen` is ever embedded outside that root, callers will need to inject navigation explicitly — there is no such use today.

PR #385, opened 2026-04-14
Align scheduled backup timestamps and log export failures

kipyin:
## Headline

Scheduled folder backups now use one consistent moment for the export, filename, and cleanup, and export failures are logged for debugging.

## User impact

Backup file names and retention cleanup now match the same point in time as the export, which avoids rare mismatches if the clock advances between steps. If a scheduled export fails while writing the temporary file, developers can see the error in the system log instead of only seeing a generic in-app failure.

## What changed

The runner captures a single date when a scheduled backup starts and passes it through the temp export, archive naming, and folder pruning so those steps do not each call the current time separately.

Failed exports to the temporary file are now recorded with an error log that includes the underlying error description to make diagnosing backup issues easier.

## Verification

On macOS with Xcode, run `grace ci` (or at least `grace test` if that matches your workflow) to confirm the project still builds and tests pass. Optionally trigger a scheduled backup path and confirm Console logs include details if the temp export fails.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Settings/Services/ScheduledBackupRunner.swift
+++ b/GraceNotes/GraceNotes/Features/Settings/Services/ScheduledBackupRunner.swift
@@ -54,7 +54,8 @@ enum ScheduledBackupRunner {
 
     /// Export to a temp file, copy into the backup folder, then delete the temp file — all on the same background task.
     private static func performScheduledExport(modelContainer: ModelContainer) -> ScheduledExportResult {
-        let exportResult = exportToTemporaryFile(modelContainer: modelContainer)
+        let exportDate = Date()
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

This update tightens scheduled backups by capturing a single **export** timestamp at the start of the run and reusing it for the archive export, the backup filename, and retention pruning, so the file name and “what counts as now” for cleanup stay consistent instead of drifting across separate `Date()` calls. It also passes that same moment into the export service and adds a clear error log when writing the temporary export fails, which should make rare backup failures easier to diagnose from device logs. Validation was not run in this environment; please run `grace test` or `grace ci` on a Mac with Xcode before merging.

PR #386, opened 2026-04-14
Make journal export filenames unique per export

kipyin:
## Headline

Each data export now gets a distinct temporary file name so rapid repeats cannot collide.

## User impact

Back-to-back exports in the same second no longer risk overwriting the previous file in the temporary directory, which avoids confusing share-sheet behavior or lost files when someone exports twice quickly. Reliability of the export flow improves without changing the archive format or what gets exported.

## What changed

The export service still writes a JSON archive under the system temporary directory with the same timestamp prefix for readability. The filename now includes an additional unique segment so two exports that share the same second cannot resolve to the same path. No changes were made to encoding, schema, or journal content.

## Verification

On a Mac with the repo and Xcode tooling, run `grace ci` (or `grace test` if that is your usual check) from the project root to confirm lint and simulator build still pass. Manually trigger two exports in quick succession from Settings and confirm both share actions receive distinct files if you want an extra sanity check.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Settings/Services/JournalDataExportService.swift
+++ b/GraceNotes/GraceNotes/Features/Settings/Services/JournalDataExportService.swift
@@ -13,7 +13,7 @@ struct JournalDataExportService {
         let entries = try context.fetch(descriptor)
         let data = try makeArchiveData(from: entries, exportedAt: now)
 
-        let filename = "grace-notes-export-\(timestampString(from: now)).json"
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

## Summary

This pass does not include any new commits or file edits, so there is nothing additional to describe in the diff. I am leaving this note for visibility because there was no reviewer feedback excerpt to respond to and no list of touched files for this round. If you expected a follow-up change or a specific fix, say what you are still seeing and I will address it in the next update.

PR #387, opened 2026-04-14
Use reliable pixel scale for journal share image rendering

kipyin:
## Headline

Exported journal share images stay sharp on Retina devices when rendered off-screen.

## User impact

Share images from your journal no longer risk looking soft or pixelated when the system reports a 1× scale during off-window rendering. The export path is more dependable across devices and sharing flows.

## What changed

Journal share cards are drawn with SwiftUI’s ImageRenderer outside a normal window. In that situation, UITraitCollection’s display scale can be missing or stuck at 1× even on Retina screens, so the old fallback could pick the wrong value and under-scale the bitmap.

The renderer now chooses the best positive scale between the current trait collection and the main screen, and uses a sensible default only if both are unusable. A short comment documents why this matters for off-window rendering.

## Verification

On a Mac with Xcode and the iOS Simulator, run `grace ci` (or `grace test` with the project’s usual destination) to confirm the app builds and tests pass. Manually share a journal entry and confirm the generated image looks crisp on a Retina simulator or device.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Services/JournalShareRenderer.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Services/JournalShareRenderer.swift
@@ -20,8 +20,12 @@ enum JournalShareRenderer {
     /// Keep in sync with `JournalShareCardView` (`cardWidth` + twice `padding`).
     private static let proposedLayoutWidth: CGFloat = 448 + 24 * 2
 
+    /// `ImageRenderer` runs without hosting in a window; `UITraitCollection.current.displayScale` can be
+    /// unset or 1× while the device screen is Retina. Take the best positive value between trait and screen.
     private static var pixelScale: CGFloat {
-        let scale = UITraitCollection.current.displayScale
-        return scale > 0 ? scale : UIScreen.main.scale
+        let traitScale = UITraitCollection.current.displayScale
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

Thanks for the detailed review—agreed there are no blocking issues for merge.

We fixed share-sheet bitmap scaling for snapshots driven by `ImageRenderer`, which is not window-hosted. In that situation `UITraitCollection.current.displayScale` can read as **0** or **1×** even when the device is Retina, while `UIScreen.main.scale` still reflects the real pixel density. The old logic treated any trait scale **> 0** as authoritative, so it could pick **1×** and produce soft or oddly laid-out images. We now take the **maximum** of the two positive scales (with a defensive fallback if both are invalid) so we align with a reliable pixel scale for layout and rendering.

On the follow-ups: extending `JournalShareRendererTests` to assert `image.scale` (or pixel dimensions against layout width) would catch a return to 1×; I agree that is a reasonable next step. Full validation remains on macOS with Xcode and the simulator via `grace test` / `grace ci`, per the repo.

| Severity | Count / note |
|----------|----------------|
| Critical | — |
| High     | — |
| Medium   | — |
| **Blocking** | **None** |

PR #388, opened 2026-04-14
Request photo library authorization before Save to Photos

kipyin:
## Headline

Saving a shared image to Photos now asks for add access first and finishes cleanly if it is denied.

## User impact

Previously, the custom share action could try to add the image without explicitly handling Photos permission, which could lead to failed or confusing saves on first use. Users now see the system add-to-library prompt when needed, and if they decline, the activity ends with a clear failure instead of an ambiguous outcome.

## What changed

The Save to Photos share activity now calls the Photos framework to request add-only access before it runs the save. The actual save logic was moved into a small helper so authorization, main-thread UI work, and the existing change request stay in a clear order. If authorization is not granted, the activity reports completion as unsuccessful and does not attempt the save.

## Verification

On a Mac with Xcode and the iOS Simulator (or a device), open the share sheet, choose Save to Photos, and confirm the permission prompt appears when needed and that saving succeeds or fails predictably. From the repo root, run `grace ci` to match CI (lint plus simulator build).

## Risk

Medium

## Touch class

`ui-ux`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Views/SaveToPhotosActivity.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Views/SaveToPhotosActivity.swift
@@ -32,6 +32,19 @@ final class SaveToPhotosActivity: UIActivity {
     }
 
     override func perform() {
+        PHPhotoLibrary.requestAuthorization(for: .addOnly) { [weak self] status in
+            DispatchQueue.main.async {
+                guard let self else { return }
+                guard status == .authorized else {
+                    self.activityDidFinish(false)
+                    return
+                }
+                self.saveImageToPhotoLibrary()
+            }
+        }
+    }
+
+    private func saveImageToPhotoLibrary() {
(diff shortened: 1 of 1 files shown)


kipyin:
Thanks for the deep review. I addressed the medium note by finishing the activity whenever the add-only authorization callback runs: denied, restricted, or not-yet-authorized paths now call `activityDidFinish(false)` on the main queue, and the authorization handler keeps a strong reference to the activity for the duration of that work so we do not drop the finish call if the instance were only weakly held. The `performChanges` completion still uses a weak capture to avoid extending lifetime past the save path; if we see confusing failures in the field, we can add targeted logging of the Photos error when `success` is false without logging image data. Automated tests still do not cover the full authorization matrix; validating denied/restricted/first-prompt behavior remains best on device or manual QA, and we can revisit user-visible copy for denied access in a follow-up if product wants more than the system sheet.

PR #389, opened 2026-04-14
Validate backup folder path is a directory

kipyin:
## Headline

Backup location resolution now fails with a clear error when a file blocks the Grace Notes Backup folder.

## User impact

If something already exists at the expected backup path but it is a file (not a folder), the app no longer treats that path like a usable backup folder. Callers can surface a reliable failure instead of odd or inconsistent filesystem behavior when backing up.

## What changed

The resolver used to call only whether a path existed before creating a directory, without distinguishing a file from a folder at the same location. That could let a mis-resolved path slip through or produce confusing behavior when backups were written.

Resolution now uses one helper that creates the directory when missing, returns the URL when the path is already a folder, and throws a dedicated error when the path exists but is not a directory. Both the direct “Grace Notes Backup” selection and the child-under-the-picked-folder case follow the same rules.

## Verification

On a Mac with Xcode, run `grace ci` (or `grace test`) from the repo root. Confirm `BackupFolderPickerResolutionTests` pass, including coverage for the non-directory path case.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Settings/Services/BackupFolderPickerResolution.swift
+++ b/GraceNotes/GraceNotes/Features/Settings/Services/BackupFolderPickerResolution.swift
@@ -4,21 +4,32 @@ enum BackupFolderPickerResolution {
     /// On-disk folder name (not localized—consistent across locales for iCloud Drive / Files).
     static let subfolderName = "Grace Notes Backup"
 
+    enum ResolutionError: Error {
+        /// A file (or other non-directory) already occupies the resolved path.
+        case pathComponentIsNotDirectory(URL)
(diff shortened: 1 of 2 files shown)


kipyin:
This pass adds a unit test that covers the edge case where the reserved backup subfolder name already exists on disk as a **file** instead of a folder. The test asserts that resolving the picked parent folder fails with `pathComponentIsNotDirectory` and points at that blocking path, so the behavior stays locked in after the review feedback. Existing tests already covered creating the subfolder under a parent and avoiding double-nesting when the user picks the subfolder directly.

PR #390, opened 2026-04-14
Lock journal share composer while export runs and fix share state

kipyin:
## Headline

Prevents edits and repeat shares while the share card image is being built and handed off.

## User impact

Tapping Share could previously leave preview controls active while rendering ran, so someone could change redactions or toggles mid-export or tap Share again. The composer now freezes those controls during the commit, and failed image creation clears the busy state so you can adjust settings and try again.

## What changed

The scrollable composer content is disabled whenever a share commit is in progress, matching the toolbar buttons and sheet dismiss behavior. The share commit path is marked to run on the main actor, and the old pattern that always reset the in-progress flag when the function returned was replaced: rendering failures explicitly clear the flag before showing the error, while a successful handoff leaves the surface locked until the flow ends (typically when the sheet is dismissed).

## Verification

On a Mac with the project toolchain, run `grace ci` (or `grace test` if you only need tests). In the simulator, open the journal share composer, tap Share, and confirm preview chips, toggles, and the scroll area do not respond until the sheet finishes or you cancel; trigger a render failure if possible and confirm you can edit and share again after dismissing the alert.

## Risk

Medium

## Touch class

`ui-ux`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Views/JournalShareComposerView.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Views/JournalShareComposerView.swift
@@ -42,6 +42,7 @@ struct JournalShareComposerView: View {
                 }
                 .padding(.horizontal, 20)
                 .padding(.vertical, 16)
+                .disabled(isShareCommitInProgress)
             }
             .scrollContentBackground(.hidden)
             .background(AppTheme.settingsBackground.ignoresSafeArea())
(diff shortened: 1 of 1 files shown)


kipyin:
Thanks for the careful review. I moved the “commit in progress” lock to the `ScrollView` with `scrollDisabled` and `allowsHitTesting`, so the whole scroll area stays frozen during share—not just the inner controls, which `disabled` on the `VStack` did not fully guarantee. After a successful render, `isShareCommitInProgress` is now cleared when `onShare` returns, so the composer does not depend on immediate sheet dismissal to recover; render failure still clears the flag and shows the error alert. `interactiveDismissDisabled` and disabled toolbar buttons keep the same guard against double taps and swipe-to-dismiss while work is in flight.

PR #391, opened 2026-04-14
Validate backup parent path is a directory before creating subfolder

kipyin:
## Headline

Choosing a backup location now fails fast with a clear error when a file blocks the folder path.

## User impact

If you pick a location where a file (not a folder) sits on the path to “Grace Notes Backup,” the app surfaces the existing “not a directory” resolution error instead of hitting a murkier filesystem failure when it tries to create the backup folder. Backup setup stays predictable and easier to recover from by picking a real folder.

## What changed

Backup folder resolution used to append the “Grace Notes Backup” subfolder and create it without first checking that the user-selected parent path, when it already exists on disk, is actually a directory. A file at that path could lead to confusing or unreliable behavior when creating the nested folder.

The change introduces a small helper that, when something already exists at a path, requires it to be a directory; otherwise it throws the same resolution error used elsewhere. That check runs for the parent path before appending the subfolder name, and the helper is shared with directory creation so the directory-vs-file logic lives in one place.

## Verification

On a Mac with the repo’s Xcode setup, run `swiftlint lint` from the repo root and `grace test` (or `grace ci` for the default lint-and-build profile). Confirm unit coverage for `BackupFolderPickerResolution` still passes, including cases where the picked path is a file or a non-directory.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Settings/Services/BackupFolderPickerResolution.swift
+++ b/GraceNotes/GraceNotes/Features/Settings/Services/BackupFolderPickerResolution.swift
@@ -17,16 +17,24 @@ enum BackupFolderPickerResolution {
         if standardized.lastPathComponent == subfolderName {
             return try ensureDirectoryExistsOrCreate(at: standardized, fileManager: fileManager)
         }
+        try requireDirectoryIfExists(at: standardized, fileManager: fileManager)
         let child = standardized.appendingPathComponent(subfolderName, isDirectory: true)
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

No `GraceNotes/**/*.swift` paths in the PR diff for automated edits; review feedback was noted.

Digest (for context):

(No review digest text.)

PR #392, opened 2026-04-14
Serialize scheduled backup runs to prevent overlapping exports

kipyin:
## Headline

Stops rapid foreground switches from stacking duplicate scheduled backups while a run is still finishing.

## User impact

When you bounce in and out of the app quickly, Grace Notes can trigger multiple scheduled backup tasks before the previous run records completion. That overlap wastes battery and storage, stresses SwiftData and disk I/O, and can produce confusing backup history. This change keeps scheduled exports to one at a time so backups stay predictable and reliable.

## What changed

The scheduled backup path now uses a small single-flight gate (a lock-backed “only one in flight” flag) around the work that runs after the folder bookmark is valid. If another activation tries to start a backup while one is already exporting and copying, the extra attempt exits early instead of launching a second detached export. The gate clears when the in-flight run finishes, whether it succeeds or fails, so the next due window can run normally.

## Verification

On a Mac with Xcode and the project’s dev tooling installed, run `grace ci` from the repo root (or `grace test` with your usual simulator destination) to lint and build the GraceNotes scheme.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Settings/Services/ScheduledBackupRunner.swift
+++ b/GraceNotes/GraceNotes/Features/Settings/Services/ScheduledBackupRunner.swift
@@ -7,6 +7,29 @@ private let scheduledBackupLogger = Logger(
     category: "ScheduledBackup"
 )
 
+/// Prevents overlapping scheduled backup runs when the app becomes active repeatedly before
+/// `lastRunAt` is updated (each transition spawns a new `Task` in ``GraceNotesApp``).
+private final class ScheduledBackupSingleFlight: @unchecked Sendable {
+    private let lock = NSLock()
+    private var inFlight = false
+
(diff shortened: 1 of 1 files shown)


kipyin:
Thanks for the thorough review.

We updated `end()` to use `defer { lock.unlock() }` so unlocking matches `tryBegin()` and stays correct if the body gains early returns later.

We moved the single-flight release into the detached export work so `inFlight` clears when `performScheduledExport` finishes, not when the outer `async` task is cancelled—this keeps the gate aligned with actual export lifetime and avoids overlap if a hosting `Task` is cancelled while `await Task.detached { … }.value` is still in flight.

We did not add automated tests for cancellation/re-entrancy in this pass; that remains a good follow-up if we want CI coverage for the coordinator.

PR #393, opened 2026-04-14
Align custom past-statistics window with validated selection and safe date bounds

kipyin:
## Headline

Past statistics no longer risk a crashing or empty date range when a custom lookback resolves oddly versus today.

## User impact

Review and statistics that depend on your chosen “past interval” stay reliable instead of failing on edge cases where the computed window does not line up cleanly with the reference day. Custom windows now behave consistently with how “all journal history” already capped awkward dates.

## What changed

The history-range helper now works from a single validated copy of your interval (mode, quantity, unit), so quantity always follows the same clamping rules as everywhere else instead of re-applying limits in one branch.

For custom intervals, the start date is now capped so it is never after the reference day, matching the safety already used for “all entries.” That keeps the half-open day range well-formed when calendar math or unusual dates would otherwise put the start after “today.”

## Verification

On a Mac with the repo’s dev setup, run `grace ci` (or `grace test` with the project’s default simulator destination) so lint and the GraceNotes build complete successfully.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Services/PastStatisticsIntervalPreference.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Services/PastStatisticsIntervalPreference.swift
@@ -46,11 +46,12 @@ struct PastStatisticsIntervalSelection: Codable, Equatable, Hashable, Sendable {
         calendar: Calendar,
         allEntries: [Journal]
     ) -> Range<Date> {
+        let selection = validated
         let refStart = calendar.startOfDay(for: referenceDate)
         guard let endExclusive = calendar.date(byAdding: .day, value: 1, to: refStart) else {
             return refStart..<refStart
         }
-        switch mode {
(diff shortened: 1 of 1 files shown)


kipyin:
Added **`test_resolvedHistoryRange_customMode_outOfBoundsQuantity_clampsAndProducesWellFormedRange`** in `PastStatisticsIntervalPreferenceTests.swift` so **`.custom`** is covered the same way **`.all`** already was for history ranges. It builds a fixed UTC calendar and reference day, then calls `resolvedHistoryRange` with **quantity 0** and **1000** (weeks) and checks that **0 clamps like 1**, **1000 clamps like 999**, and each result is a **valid half-open range** (`lowerBound < upperBound`). That matches the review ask to exercise **out-of-bounds quantities** and assert **clamped behavior** and a **well-formed** range for the custom path.

PR #395, opened 2026-04-14
Harden backup folder list: scoped access helper and selection sync

kipyin:
## Headline

Backup exports list stays trustworthy when the folder contents refresh or the list reloads.

## User impact

People managing scheduled backup files in Settings see a list that stays in sync with what is actually on disk. Stale or invalid selections are dropped after a refresh, so delete confirmations and actions match real files and are less likely to target the wrong thing or behave oddly after the list changes.

## What changed

Loading the backup folder file list now uses the same security-scoped folder access helper as delete, instead of hand-rolled start/stop access around the folder URL. That keeps access rules in one place and reduces the chance of inconsistent handling between listing and deleting.

When the file array updates, the multi-select set is intersected with the paths that still exist in the new list. Empty lists still exit selection mode; non-empty refreshes prune selections for files that disappeared without wiping the whole set when some files remain.

## Verification

On a Mac with Xcode and the project’s grace CLI installed, run `grace ci` (or `grace test` with the default simulator destination from gracenotes-dev.toml). In the app, open Settings → the scheduled backup folder file list, enter multi-select, refresh the list (or change folder contents), and confirm selection and delete flows still match visible files.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Settings/ImportExportSettingsSupport.swift
+++ b/GraceNotes/GraceNotes/Features/Settings/ImportExportSettingsSupport.swift
@@ -159,9 +159,10 @@ struct BackupFolderImportFileListView: View {
             load()
         }
         .onChange(of: files) { _, newFiles in
+            let validPaths = Set(newFiles.map(\.path))
+            selection = selection.intersection(validPaths)
             if newFiles.isEmpty {
                 isSelecting = false
-                selection.removeAll()
             }
         }
     }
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

Thanks for the detailed review.

We updated `load()` to use the same security-scoped bookmark flow as scheduled backups and delete: resolve the bookmark, call `startAccessingSecurityScopedResource()`, and always pair it with `stopAccessing…` in a `defer`, routing failures through `ScheduledBackupError.securityScopeDenied` so denied scope behaves consistently across paths. On a successful reload we rebuild selection from the fresh file list (`Set(newFiles.map(\.path))`) so edit-mode tags do not reference files that are no longer present if the directory contents change between loads; an empty listing still clears selection because intersecting with an empty set drops everything. We did not change the behavior on `load()` failure: `files` stays as before and `listError` drives the error UI, matching the previous contract. We agree a quick **macOS Simulator** pass in Settings → Import/Export → backup folder (picker, revoke/regrant access, multi-select delete while the list refreshes) is the right follow-up since that is not exercised in Linux CI.

PR #396, opened 2026-04-14
Validate curated review theme labels in DEBUG builds

kipyin:
## Headline

Catch missing theme display labels before they ship and break localized review themes.

## User impact

When a curated alias points at a canonical theme that has no display label, the app can fall back to generic title-casing and skip proper `review.theme.*` strings, so labels look wrong or inconsistent. This change does not alter runtime behavior in release; it only fails fast in development when the map is incomplete.

## What changed

The default curated theme map is now built inside a small initializer closure so we can run a DEBUG-only consistency check before the static value is exposed. A new assertion walks every canonical target referenced by global aliases, section overrides, and cross-language aliases and requires each one to exist in `canonicalDisplayLabels`, crashing with a clear list of missing keys if not. That keeps the mapping table and localization story aligned as editors add or tweak aliases.

## Verification

On a Mac with Xcode, run `swiftlint lint` from the repo root and `grace ci` (or `grace test` with the project’s default simulator destination) to confirm the app still builds and tests pass; in DEBUG, launching code paths that touch `ReviewCuratedThemeMap.defaultMap` will trip the assertion if any canonical is missing a label.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Services/ReviewCuratedThemeMap.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Services/ReviewCuratedThemeMap.swift
@@ -8,159 +8,189 @@ struct ReviewCuratedThemeMap {
     let crossLanguageAliases: [String: String]
     let canonicalDisplayLabels: [String: String]
 
-    static let defaultMap = ReviewCuratedThemeMap(
-        globalAliases: [
-            "sleep": "sleep",
-            "better sleep": "sleep",
-            "slept": "sleep",
-            "rest": "rest",
-            "recover": "rest",
-            "recovery": "rest",
-            "downtime": "rest",
-            "quiet": "quiet time",
-            "quiet time": "quiet time",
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

*(No review digest text.)*

## Files touched in this pass

- `GraceNotes/GraceNotes/Features/Journal/Services/ReviewCuratedThemeMap.swift`

This update refines the default **review curated theme map** (aliases, section overrides, banned and penalized concepts, cross-language aliases, and canonical display labels) and builds `defaultMap` through a typed closure so the structure stays explicit. In **Debug** builds we now assert that every canonical target produced by those mappings has a matching entry in `canonicalDisplayLabels`, which catches gaps early instead of falling back to naive title-casing and missing `review.theme.*` localization. **Validation:** please run `grace ci` or `grace test` on a Mac with Xcode and the iOS Simulator; this environment cannot compile the iOS app.

PR #401, opened 2026-04-14
Sentry: refine ShareSheet.swift

kipyin:
## Headline

Automated refinement from `grace sentry`.

## User impact

Maintainers get a small scoped change with CI preflight before review.

## What changed

Updates `GraceNotes/GraceNotes/Features/Journal/Views/ShareSheet.swift` in a single automated pass (see diff on the PR).

## Verification

`grace ci` on macOS using the configured sentry CI profile.

## Risk

Medium

## Touch class

`ui-ux`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Views/ShareSheet.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Views/ShareSheet.swift
@@ -12,11 +12,34 @@ struct ShareSheet: UIViewControllerRepresentable {
     var applicationActivities: [UIActivity]?
 
     func makeUIViewController(context: Context) -> UIActivityViewController {
-        UIActivityViewController(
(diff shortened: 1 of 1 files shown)


kipyin:
**Response to review**

Popover anchoring no longer depends on finding a key window during `makeUIViewController`. A small `UIActivityViewController` subclass reapplies popover `sourceView` / `sourceRect` in `viewDidLayoutSubviews` and `viewDidAppear`, when the activity view is in the hierarchy and has non-zero bounds, with an extra main-queue pass after creation for the first frame. The anchor uses the activity controller’s own view instead of a global key-window search. Empty activity items are guarded with `precondition` so Release builds still fail fast, and `updateUIViewController` now refreshes popover configuration if the representable updates. We have not added UI tests for iPad regular-width share or Settings export; those remain good manual or follow-up automation targets.

PR #404, opened 2026-04-14
Align review calendar VoiceOver hints with day state

kipyin:
## Headline

VoiceOver now describes each drill-down day button by what actually happens when you open it.

## User impact

People using VoiceOver were told every tappable day would open an existing entry, even on days with no journal yet. That mismatch made the Past review calendar harder to trust and navigate. The hints now match empty days versus days that already have writing.

## What changed

The review history drill-down calendar grid no longer applies one generic “open entry” VoiceOver hint to every in-range day button. A small helper picks the hint from the day’s disposition: days with no journal yet describe opening the day to write, while matched or otherwise journal-backed days describe opening what you wrote that day, and out-of-range cases stay non-interactive as before. An unused SwiftData import was removed because this view does not use SwiftData types directly, which keeps the module surface clear and avoids implying a dependency that is not there.

## Verification

On a Mac with Xcode and the iOS Simulator, run `grace ci` (or `grace test` with the project’s default destination). For accessibility, enable VoiceOver on the simulator, open the relevant Past review drill-down calendar, and confirm empty in-range days announce a write-oriented hint while days with journal content announce an entry-oriented hint.

## Risk

Medium

## Touch class

`ui-ux`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Views/ReviewHistoryDrilldownCalendarGrid.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Views/ReviewHistoryDrilldownCalendarGrid.swift
@@ -1,5 +1,4 @@
 import SwiftUI
-import SwiftData
 
 // MARK: - Section strip (read-only)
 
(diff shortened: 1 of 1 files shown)


kipyin:
Addressed the review note about the unreachable `.outsideHistoryWindow` branch in `dayCellButtonAccessibilityHint`: that path is only called from the tappable `Button` branch in `dayCell`, while outside-window days render on the non-button path, so the old empty-string case was dead code. Replaced it with a `fatalError` and a short doc comment so a future layout change that accidentally routes outside-window cells through the button path fails fast in development instead of silently shipping a wrong hint. Left the switch exhaustive and did not change VoiceOver copy or behavior for valid flows. Validation is still `grace test` / `grace ci` on a Mac with Xcode and the iOS Simulator.

PR #405, opened 2026-04-14
Only mark today loaded after successful journal load

kipyin:
## Headline

Opening today’s journal can recover when the first load fails instead of getting stuck.

## User impact

If loading today’s entry failed (for example a SwiftData fetch error), the app used to treat today as already loaded anyway, so you could see an empty or wrong screen with no second chance to load. We now flip that flag only after a successful load so the next attempt can try again.

## What changed

Previously, “today” was marked as loaded before the repository finished, so a failed load still blocked retries.

`loadEntry` now returns whether the load completed successfully: true when an existing journal is hydrated or a new one is inserted, and false when the fetch fails and the localized load error is set. `loadTodayIfNeeded` sets the loaded flag only when that succeeds. The result is marked discardable so other call sites that ignore the return value stay clean. A small optional-chaining tweak in the autosave sink matches the surrounding style.

## Verification

On a Mac with Xcode and the iOS Simulator, from the repo root run `grace ci` (or `grace test` with your usual destination). Manually open the journal after inducing or observing a load failure to confirm a later appearance can load today successfully.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/ViewModels/JournalViewModel.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/ViewModels/JournalViewModel.swift
@@ -63,11 +63,13 @@ final class JournalViewModel {
 
     func loadTodayIfNeeded(using context: ModelContext) {
         guard !hasLoadedToday else { return }
-        hasLoadedToday = true
-        loadEntry(for: nowProvider(), using: context)
+        if loadEntry(for: nowProvider(), using: context) {
+            hasLoadedToday = true
+        }
     }
 
-    func loadEntry(for date: Date, using context: ModelContext) {
+    @discardableResult
+    func loadEntry(for date: Date, using context: ModelContext) -> Bool {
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

**Update:** `hasLoadedToday` is now set only after `loadEntry` completes successfully, so a failed fetch no longer marks today as “loaded” and blocks a later retry. That matches the failure mode the review called out: an early guard flip paired with a miss path that could still insert an unsaved journal. **Follow-up:** I agree the retry contract deserves a unit test in `JournalViewModelTests` with a stub repository (fail once, then succeed) so CI catches any regression back to “set guard before load.” I have not added that test in this reply; I will add it in a follow-up commit if it is not already on the branch. **Note on validation:** `GraceNotesTests` run via Xcode/`grace test` on macOS; Linux agents still cannot execute that target per `AGENTS.md`. The `refreshTodayIfStale` / `loadEntry` return-value and localized error-key points in the review are noted as pre-existing follow-ups, not introduced by this change.

PR #406, opened 2026-04-14
Derive hex SwiftUI colors from UIColor for consistent conversion

kipyin:
## Headline

Theme colors built from hex values now share one UIKit-backed conversion path so the app renders them the same everywhere.

## User impact

Hex-based Warm Paper and other theme colors should look identical whether they are used through SwiftUI or underlying UIKit APIs, with less risk of tiny shifts from duplicate math. That keeps backgrounds, accents, and borders visually stable across screens and modes.

## What changed

The SwiftUI helper that turns a 24-bit hex value into a color no longer computes red, green, and blue with its own floating-point math. Instead it builds a UIKit color from those components and wraps it in SwiftUI, so there is a single definition of how hex maps to sRGB channels.

The private UIColor hex initializer was moved above the Color helper so that path is defined before use, without changing which hex values the theme exposes.

## Verification

On a Mac with Xcode and the project’s grace CLI installed, run `grace ci` (or `grace test` with the usual simulator destination from gracenotes-dev.toml) to confirm the GraceNotes scheme still builds and tests pass. Optionally spot-check a few hex-driven surfaces (e.g. journal chrome and review accents) in light and dark mode in the Simulator.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/DesignSystem/Theme.swift
+++ b/GraceNotes/GraceNotes/DesignSystem/Theme.swift
@@ -349,12 +349,18 @@ struct WarmPaperPressStyle: ButtonStyle {
 
 // MARK: - Color Hex Extension
 
+private extension UIColor {
+    convenience init(hex: UInt) {
+        let red = CGFloat((hex >> 16) & 0xFF) / 255
+        let green = CGFloat((hex >> 8) & 0xFF) / 255
+        let blue = CGFloat(hex & 0xFF) / 255
+        self.init(red: red, green: green, blue: blue, alpha: 1)
+    }
+}
+
 private extension Color {
     init(hex: UInt) {
-        let red = Double((hex >> 16) & 0xFF) / 255
-        let green = Double((hex >> 8) & 0xFF) / 255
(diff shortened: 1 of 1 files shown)


kipyin:
**Update:** In this branch there is a single `Color(hex:)` / `UIColor(hex:)` implementation in `Theme.swift`; `ShareCardStyle` and `ShareCardSurface` only call that initializer—they do not define their own hex conversion, so share cards already use the same UIKit-backed path as `AppTheme`.

This change moves the authoritative conversion to `UIColor(red:green:blue:alpha:)` and has `Color` wrap that, so design-system colors and any `Color(hex:)` usage stay aligned without duplicating bit-masking math.

We should tighten the PR description so it does not imply a repo-wide audit of every RGB construction (for example `TodayJournalPalette` still has its own path where that was intentional).

**Manual check:** A quick pass on a device or simulator for a couple of themed screens and a share-card export is still the right sanity check; CI cannot prove pixel-level parity for SwiftUI colors.

PR #407, opened 2026-04-14
Optimize Past statistics section chip counts with single-pass day lookup

kipyin:
## Headline

Past statistics drill-down chip math stays the same but scales better when you have lots of journal history.

## User impact

Computing how many section chips to show for each matched day used to rescan your contributing entries for every day, which could add noticeable work on large histories. The app now picks the newest contributing entry per calendar day once, then reads counts from that map, so scrolling and stats stay responsive without changing what you see.

## What changed

The per-day chip helper for review statistics previously walked the newest-first contributing list with a linear search for each matched day, which could multiply work when many days matched. It now makes one forward pass over that list, recording the first (newest) journal per start-of-day key, reserves dictionary capacity, and uses those lookups when iterating matched days. Behavior matches the old “newest entry wins when several fall on the same day” rule that the drill-down tests lock in.

## Verification

On a Mac with Xcode and the project’s dev tooling installed, run `grace ci` (or `grace test` with the repo’s default simulator destination) to lint and build; exercise Past statistics with sections that have multiple entries on the same day to confirm chip counts still match expectations.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Services/ReviewHistoryWindowing.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Services/ReviewHistoryWindowing.swift
@@ -98,11 +98,20 @@ enum ReviewHistoryWindowing {
         contributingEntriesNewestFirst: [Journal],
         calendar: Calendar
     ) -> [Date: Int] {
+        // One pass: first row per calendar day matches `first(where:)` (see drill-down tests).
+        var firstContributingByDay: [Date: Journal] = [:]
+        firstContributingByDay.reserveCapacity(contributingEntriesNewestFirst.count)
+        for entry in contributingEntriesNewestFirst {
+            let day = calendar.startOfDay(for: entry.entryDate)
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

**Response to deep review**

We refactored `sectionChipCountByMatchedDays` so it walks `contributingEntriesNewestFirst` once, records the first journal seen for each calendar day (which is still the newest contributor when the list is newest-first), then looks up those journals for each `matchingDayStarts` key instead of scanning the full entry list per day. That keeps the same chip counts and the same “skip days with no contributing entry” behavior as before, with lower asymptotic cost when there are many days and entries. We did not add an on-device benchmark; the gain is expected on large histories rather than something we measured here. The existing drill-down test that locks “first contributing row per day” remains the contract, and callers should keep passing `matchingDayStarts` as `startOfDay` values from the same `Calendar` used for the entries, as `ReviewHistoryDrilldownSheets` already does.

PR #408, opened 2026-04-14
Stabilize journal backup export order with deterministic tie-breakers

kipyin:
## Headline

Backups list same-day journals in a stable, repeatable order instead of an arbitrary one.

## User impact

If you have more than one journal on the same calendar day, exported JSON used to order those rows in a way that could vary between exports. That made diffs, scripts, and side-by-side comparisons noisy and could confuse anyone relying on a fixed sequence. Exports now use a clear, consistent order so backups are predictable and easier to trust.

## What changed

Journal export used to sort only by the journal’s calendar day. When two records shared that day, the relative order was not fully defined, so SwiftData fetches and simple array sorts could shuffle ties.

The export path now breaks ties the same way in both places: after the calendar day, it sorts by when the journal was created, and if those still match, by stable journal identity. Fetching from the store uses matching sort keys so read order aligns with the archive builder when entries are passed in from a query.

## Verification

On a Mac with Xcode, run `grace ci` (or `grace test`) from the repo root to confirm the project still builds and tests pass.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Settings/Services/JournalDataExportService.swift
+++ b/GraceNotes/GraceNotes/Features/Settings/Services/JournalDataExportService.swift
@@ -8,7 +8,11 @@ struct JournalDataExportService {
         fileManager: FileManager = .default
     ) throws -> URL {
         let descriptor = FetchDescriptor<Journal>(
-            sortBy: [SortDescriptor(\.entryDate, order: .forward)]
+            sortBy: [
+                SortDescriptor(\.entryDate, order: .forward),
+                SortDescriptor(\.createdAt, order: .forward),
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

Journal exports now use a **stable** order: journals are sorted by `entryDate`, then by `createdAt`, then by `id` when building the export archive and when fetching from SwiftData. That way day-level ties don’t produce random ordering across runs or devices, so backups and shared files stay consistent and easier to diff or compare. Please run **`grace test`** or **`grace ci`** on your Mac with the iOS Simulator to confirm everything still passes in your environment.

PR #409, opened 2026-04-14
Normalize weekly review day activity to calendar-day boundaries

kipyin:
## Headline

Weekly review’s per-day activity strip now lines up with real calendar days and your saved entries.

## User impact

If the review window’s dates weren’t exactly at local midnight, day rows could carry a mismatched timestamp while lookups used start-of-day keys, which could skew how completion and “had an entry” line up in the weekly summary. The change makes each row’s date and the loop’s cursor the same normalized calendar day so the strip stays consistent and trustworthy.

## What changed

The code that walks the current review week and builds `ReviewDayActivity` no longer carries a raw `Date` through the loop while comparing against dictionary keys built from `startOfDay`. It initializes from the period’s lower bound normalized to the start of that day, uses that same value for each row’s `date`, and steps forward by adding a day then re-normalizing to start-of-day. If advancing the day fails, the loop stops cleanly instead of forcing a fallback that could hide calendar edge cases.

## Verification

On a Mac with Xcode, run `grace ci` (or `grace test` with the project’s usual simulator destination) to confirm the GraceNotes scheme still builds and tests pass.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Services/WeeklyReviewAggregates.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Services/WeeklyReviewAggregates.swift
@@ -435,13 +435,12 @@ private extension WeeklyReviewAggregatesBuilder {
                 .map { calendar.startOfDay(for: $0.entryDate) }
         )
         var activity: [ReviewDayActivity] = []
-        var day = currentPeriod.lowerBound
-        while day < currentPeriod.upperBound {
-            let dayStart = calendar.startOfDay(for: day)
+        var dayStart = calendar.startOfDay(for: currentPeriod.lowerBound)
+        while dayStart < currentPeriod.upperBound {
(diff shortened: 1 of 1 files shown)


kipyin:
This update tightens weekly review aggregation so a `currentPeriod` whose lower bound is not midnight (for example, noon from cached or legacy data) still lines up with `calendar.startOfDay`. Rhythm history now walks days from a normalized start through the week end, and `buildRhythmHistory` caps how many calendar days we materialize (731, about two years) so very old journals cannot grow an unbounded `ReviewDayActivity` array and overload SwiftUI. New tests cover the skewed-period case (activity matches the aligned period, dates are start-of-day), the horizon cap when the earliest entry is decades ago, and that multi-month spans still work when under the cap. We have not changed theme locale threading, completion tie-breaking, or virtualization in this pass; those remain reasonable follow-ups if we want to harden mixed-locale themes or UI scale further.

PR #411, opened 2026-04-14
Deduplicate merge conflict days during journal import

kipyin:
## Headline

Merge import conflicts now list each affected day once, not repeated dates.

## User impact

When a merge import finds overlapping days that disagree with what is on the device, the app should name each calendar day only once. Duplicate entries made conflict lists noisy, harder to scan, and could exaggerate how many decisions you need to make.

## What changed

Conflict detection for merge imports used to append every conflicting day into an array, which could repeat the same calendar day. The service now records conflicting days in a set keyed by start-of-day, then returns them sorted, and the comment clarifies that these are unique calendar days. Behavior for a single pass over deduplicated entries stays the same, but the result is guaranteed unique and matches the documented intent.

## Verification

On a Mac with Xcode and the project’s grace CLI installed, run `grace ci` (or `grace test` with the repo’s default simulator destination) to lint and build the GraceNotes scheme.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Settings/Services/JournalDataImportService.swift
+++ b/GraceNotes/GraceNotes/Features/Settings/Services/JournalDataImportService.swift
@@ -167,14 +167,14 @@ struct JournalDataImportService {
         return (inserted, updated)
     }
 
-    /// Day starts (start-of-day) where merge mode would need a conflict decision.
+    /// Unique calendar day starts (start-of-day) where merge mode would need a conflict decision.
     func mergeConflictDayStarts(
         entries: [JournalDataExportEntry],
         context: ModelContext,
         calendar: Calendar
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

Updated merge conflict detection so calendar days are deduplicated with a `Set` of day-start instants, then sorted before returning—so callers see each conflict day once in a stable order. `Date` hashing matches `calendar.startOfDay(for:)`, so membership stays consistent for the same calendar. `importData` already runs day-level deduplication upstream, so this mainly hardens `mergeConflictDayStarts` for other or future call paths. Broader import/merge behavior is unchanged; a follow-up could add a small unit test that feeds duplicate same-day entries to lock in the uniqueness guarantee if this API stays part of the public surface.

PR #413, opened 2026-04-14
Harden guided journal onboarding banners and step progression

kipyin:
## Headline

Journal onboarding stays on the right step, shows guidance only when there is real copy, and handles bad counts safely.

## User impact

First-time journal guidance is less likely to show empty or misplaced banners or get stuck on the wrong step if section counts are inconsistent. People going through the gratitude → need → person flow see clearer, more reliable coaching above the right section.

## What changed

Section guidance no longer requires a title for the first gratitude step; titles are optional in the helper and collapse to an empty banner title when absent, while messages must be non-empty so blank rows are not treated as active guidance.

A small mapping on each onboarding step names which section hosts the banner, so placement stays aligned with the linear flow without duplicating switch logic.

Counts for gratitudes, needs, and people are clamped to zero or more before deciding the next step, so negative or corrupted values do not skip ahead or mis-order the guided path.

## Verification

On a Mac with Xcode and the default simulator setup from gracenotes-dev.toml, run `grace ci` (or `grace test` with the project’s usual destination) and walk through guided journaling on Today: confirm the first gratitude banner appears with body copy and keyboard hint, then needs and people steps after 1/1/1, with no empty banner when copy is missing.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Tutorial/JournalOnboardingFlowEvaluator.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Tutorial/JournalOnboardingFlowEvaluator.swift
@@ -6,6 +6,20 @@ enum JournalOnboardingStep: Equatable {
     case person
 }
 
+private extension JournalOnboardingStep {
+    /// Section whose row hosts the onboarding banner for this linear step.
+    var bannerSection: JournalOnboardingSection {
+        switch self {
+        case .gratitude:
+            .gratitude
+        case .need:
+            .need
+        case .person:
+            .person
+        }
+    }
+}
+
 enum JournalOnboardingSection: Hashable {
     case gratitude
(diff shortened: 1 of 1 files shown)


kipyin:
Aligned `isGuidanceActive` with real coaching copy by requiring a non-empty `message`, guarding `sectionGuidance(for:)` the same way, and returning inactive onboarding from the private `presentation` helper when the message is empty (with a DEBUG assertion so bad catalog data is caught in development). Clamped section counts with `max(0, …)` so corrupted negative values do not skip ahead in the linear flow. Added unit tests for negative gratitude and need counts so that ordering and active/locked states stay covered in CI. Full validation is still `grace test` / `grace ci` on macOS with the iOS Simulator.

PR #414, opened 2026-04-14
Improve journal share image scale and bitmap validation

kipyin:
## Headline

Shared journal cards render at the right pixel density more reliably and invalid images are rejected sooner.

## User impact

When you share a journal as an image, the export should look sharp on Retina displays instead of accidentally rendering at 1×. The update also avoids treating empty backing bitmaps as a successful share.

## What changed

Share rendering used trait and main-screen scale only; off-window rendering can still report a weak scale. The renderer now picks the strongest positive scale among the current trait, any connected window scene’s screen, and the main screen, with the same safe fallback when nothing valid is available. After building the image, it also checks the Core Graphics backing size when present so zero-sized bitmaps are not returned as success.

## Verification

On a Mac with Xcode, run `grace ci` (or `grace test` per project defaults). Manually share a journal entry and confirm the image looks crisp on a Retina device or simulator and that sharing still completes when content is valid.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Services/JournalShareRenderer.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Services/JournalShareRenderer.swift
@@ -11,9 +11,12 @@ enum JournalShareRenderer {
         let cardView = JournalShareCardView(payload: payload, onLineTap: nil)
         let renderer = ImageRenderer(content: cardView)
         renderer.proposedSize = ProposedViewSize(width: Self.proposedLayoutWidth, height: nil)
-        renderer.scale = Self.pixelScale
+        renderer.scale = Self.recommendedPixelScale()
         guard let image = renderer.uiImage else { return nil }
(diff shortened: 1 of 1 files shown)


kipyin:
Thanks for the review.

We now treat a missing `cgImage` as a failed export: if `UIImage` has no bitmap backing, we return `nil` instead of skipping pixel checks and still returning the image, so share output stays bitmap-validated end-to-end.

The `recommendedPixelScale()` doc comment is updated to match the implementation: we take the maximum positive scale among the current trait collection, every connected `UIWindowScene`’s screen, and `UIScreen.main`, then default to `3` when nothing usable is reported—we are not filtering by activation state; the comment no longer implies an “active scene only” priority.

Validation on our side: `swiftlint` on the touched file; full `grace test` / `grace ci` should be run on a Mac with Xcode as usual.

PR #415, opened 2026-04-14
Harden hex colors for share cards with sRGB and 24-bit masking

kipyin:
## Headline

Share card theme colors stay predictable when hex values include extra high bits or need explicit sRGB.

## User impact

Dark share card backgrounds, text, and accents should look consistent with the intended palette instead of drifting if a hex literal ever carries alpha or other high bits. Explicit sRGB keeps exported or rendered share images aligned with the same color math across contexts.

## What changed

The private `Color(hex:)` helper used for dark share card chrome now keeps only the low 24 bits before splitting into red, green, and blue, so values shaped like `AARRGGBB` still map to the same RGB triplet as plain `RRGGBB` literals. Color construction uses the sRGB initializer so components are interpreted in a single, well-defined color space. A short doc comment documents the masking behavior for future call sites.

## Verification

On a Mac with Xcode and the project’s `grace` CLI, run `grace ci` (or `grace test` with the default simulator destination) to lint and build; optionally open the share flow and confirm dark card styles still match expectations.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Services/ShareCardSurface.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Services/ShareCardSurface.swift
@@ -111,10 +111,12 @@ private func darkCompletionChipBackground(style: ShareCardStyle) -> some View {
 }
 
 private extension Color {
+    /// 24-bit `RRGGBB`. High bits are masked so `AARRGGBB`-style literals still resolve to the same sRGB triplet.
     init(hex: UInt) {
-        let red = Double((hex >> 16) & 0xFF) / 255
-        let green = Double((hex >> 8) & 0xFF) / 255
-        let blue = Double(hex & 0xFF) / 255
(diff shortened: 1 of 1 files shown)


kipyin:
**Response to review:** Duplicated `Color` hex logic in `ShareCardSurface.swift` and `ShareCardStyle.swift` is replaced by one place: `ShareCardHexColor.swift` defines `Color.init(hexSRGB:)` with the same 24-bit mask and explicit sRGB components, and both share-card files call that initializer for their palette literals. That keeps light/dark share-card colors aligned if hex handling ever changes. The initializer name documents that it is sRGB-only opaque RGB; the doc comment still notes that only the low 24 bits are used so `AARRGGBB`-style values do not get a separate alpha path. **Verification:** Not run in this environment; please run `grace ci` on a Mac as in `AGENTS.md`.

PR #416, opened 2026-04-14
Harden Past-tab theme language resolution for huge journals

kipyin:
## Headline

Past tab theme copy picks English or Chinese more reliably and does not waste work on enormous journal text.

## User impact

Theme chip wording on the Past tab follows your journal language, but very large entries could previously trigger expensive full-text normalization. Edge cases with odd confidence scores could also mis-handle the detector output. This update keeps the experience responsive and more predictable when journals are huge or detection is ambiguous.

## What changed

Language detection used to collapse repeated whitespace across the entire journal before anything else, which could be costly for pathologically long text. Whitespace is now trimmed only at the ends of the corpus; repeated spaces and line breaks are normalized only on a bounded prefix used for analysis, so work stays proportional to what we actually feed to the recognizer and script tie-break. The classifier path now treats non-finite confidence values as failing the bar instead of comparing them naively to the threshold. Naming was clarified so the script tie-break explicitly uses the same normalized analysis sample as the language recognizer.

## Verification

On a Mac with the repo’s tooling, run `grace ci` (or `grace test` if you prefer a narrower check) to confirm the GraceNotes scheme still lints and builds. Optionally spot-check Past tab theme chips with a very long mixed-language journal and with short text to confirm English fallbacks still feel right.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Services/ReviewJournalThemeLanguageResolver.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Services/ReviewJournalThemeLanguageResolver.swift
@@ -23,8 +23,7 @@ struct ReviewJournalThemeLanguageResolver: ReviewJournalThemeLanguageResolving {
     }
 
     func resolvedDisplayLocale(forJournalCorpus corpus: String) -> Locale {
-        let collapsed = corpus.replacingOccurrences(of: "\\s+", with: " ", options: .regularExpression)
-        let trimmed = collapsed.trimmingCharacters(in: .whitespacesAndNewlines)
+        let trimmed = corpus.trimmingCharacters(in: .whitespacesAndNewlines)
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

None in this update.

## Files touched in this pass

`GraceNotes/GraceNotes/Features/Journal/Services/ReviewJournalThemeLanguageResolver.swift`

This change refines how the Past tab picks English vs. Chinese theme chip copy from your journal text. Whitespace normalization now runs only on the bounded analysis prefix (up to the grapheme cap), so huge entries no longer trigger a whole-string regex replace. The resolver trims outer whitespace first, then applies that cheaper path. It also ignores non-finite language-confidence scores before accepting `NaturalLanguage`’s dominant language, and script tie-breaking uses the same normalized analysis string throughout for consistent behavior.

PR #417, opened 2026-04-14
Normalize streak day boundaries in StreakCalculator

kipyin:
## Headline

Journal streak counts stay accurate when the calendar steps between days.

## User impact

Streak numbers could be wrong or unstable if “yesterday” and each step backward did not match the same midnight-normalized dates used everywhere else. After this change, basic and perfect streaks line up with how journal days are keyed, so what you see reflects real consecutive journaling days.

## What changed

The streak logic already stores completion per calendar day using start-of-day dates. The code now applies the same start-of-day normalization when it derives “yesterday” from today and when it walks backward day by day in the streak loop, so dictionary lookups always use the same date keys. A safety check stops the backward walk if the calendar would ever produce a non-advancing “previous” day, avoiding a runaway loop.

## Verification

On a Mac with Xcode and the repo’s grace CLI installed, run `grace ci` (or `grace test` with the project’s default simulator destination) to lint and exercise the GraceNotes scheme.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Services/StreakCalculator.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Services/StreakCalculator.swift
@@ -49,6 +49,7 @@ struct StreakCalculator {
 
     private func currentStreakLength(byDay: [Date: Bool], today: Date) -> Int {
         let yesterday = calendar.date(byAdding: .day, value: -1, to: today)
+            .map { calendar.startOfDay(for: $0) }
         let startDay: Date
         if byDay[today] == true {
             startDay = today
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

Added a `StreakCalculator` unit test that uses a Gregorian calendar in `America/Los_Angeles` with journal dates spanning a daylight saving transition, so the streak walk-back hits the same `date(byAdding: .day, …)` plus `startOfDay` normalization path as real users outside UTC. The test checks that a multi-day streak still counts correctly when completion keys are calendar day starts, which guards against regressions where the prior-day cursor would not align with those keys. Left the existing GMT-based tests in place so default setup stays fast and deterministic. Swift tests stay Xcode-only here; this case is meant to be exercised with `grace test` / `xcodebuild test` on a Mac.

PR #419, opened 2026-04-14
Tighten weekly review theme support matching for Latin and Han text

kipyin:
## Headline

Weekly review themes attach supporting evidence only for real word matches, not accidental substrings.

## User impact

Supporting lines under weekly themes (from readings and reflections) should feel more relevant to the theme label. You should see fewer unrelated snippets where a short word appeared inside another word. Chinese and Japanese theme labels are compared using normalized text on both sides, so matching stays consistent.

## What changed

The weekly review builder attaches extra support text to a theme when a moderate semantic match is found. The previous approach used naive substring checks and compared normalized support text to an unnormalized theme string. That could incorrectly link evidence when a theme word appeared inside an unrelated word (for example, rest inside forest), and it made Latin and mixed comparisons inconsistent. The update normalizes both the theme label and the support text before comparing them. For Latin script it requires whole-word or whole-phrase boundaries so substring hits inside longer words no longer count. For text that includes Han characters it uses the normalized strings for containment checks in both directions. Longer-word overlap still uses word tokens derived from the normalized theme label.

## Verification

On macOS with Xcode and the Grace Notes dev CLI installed, run `grace ci` from the repo root (or `grace test` with the default simulator destination from gracenotes-dev.toml). Confirm the GraceNotes scheme builds and automated tests pass; optionally open Weekly Review and spot-check that theme evidence looks on-topic for English and CJK journals.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Services/WeeklyReviewAggregatesBuilder+ThemeSections.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Services/WeeklyReviewAggregatesBuilder+ThemeSections.swift
@@ -278,19 +278,30 @@ extension WeeklyReviewAggregatesBuilder {
 
     func moderateSurfaceSemanticMatch(themeConcept: String, supportText: String) -> Bool {
         let normalizedSupport = textNormalizer.normalizeThemeLabel(supportText)
-        guard !normalizedSupport.isEmpty else { return false }
-        if normalizedSupport.contains(themeConcept) || themeConcept.contains(normalizedSupport) {
+        let normalizedTheme = textNormalizer.normalizeThemeLabel(themeConcept)
+        guard !normalizedSupport.isEmpty, !normalizedTheme.isEmpty else { return false }
+        if containsHanCharacters(themeConcept) {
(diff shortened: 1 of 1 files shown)


kipyin:

kipyin:
Thanks for the thorough review. We tightened weekly-review supporting-evidence matching so theme labels and reading/reflection text are compared on the same normalized strings, and we only use regex `\b` word boundaries when both sides are Latin letters, digits, and whitespace (so kana, Cyrillic, emoji-heavy text, and similar do not go through the brittle `\b` path). For those other scripts we match with `enumerateSubstrings(.byWords)`–style whole-word/phrase checks and a guarded containment fallback that still blocks obvious ASCII substring false positives like “rest” inside “forest.” The Han branch keys off both the normalized theme and the canonical concept so routing stays aligned with what we compare, with a short comment that normalization does not strip Han. We added focused `GraceNotesTests` around `moderateSurfaceSemanticMatch` for the Latin false-positive case, Latin whole-word/phrase positives, Han containment after normalization, and punctuation-only support text.

PR #421, opened 2026-04-14
Trim journal onboarding guidance and hide empty whitespace-only cards

kipyin:
## Headline

Onboarding hints no longer render blank cards or invisible lines when copy is only whitespace.

## User impact

Journal onboarding guidance treats trimmed text as the real content, so you do not get an empty-looking card or stray blank lines from spaces or newlines. Title-only or secondary-line-only hints can appear when the main message is empty after trimming, which matches how those strings are meant to be used.

## What changed

The guidance view now trims leading and trailing whitespace from the title, main message, and optional second line. It shows the card only when at least one of those trimmed values is non-empty, instead of keying visibility solely on the raw main message. Each line is drawn only when it has non-empty trimmed text, and an all-whitespace optional second line is dropped so it does not add noise.

## Verification

On a Mac with the repo and Xcode, run `grace ci` (or `grace test`) from the project root to confirm the app still builds and tests pass; in the simulator, open journal onboarding flows that use this view and confirm cards with only spaces do not appear, while title-only or hint-only content still shows as expected.

## Risk

Medium

## Touch class

`ui-ux`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Tutorial/JournalOnboardingGuidanceView.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Tutorial/JournalOnboardingGuidanceView.swift
@@ -13,25 +13,45 @@ struct JournalOnboardingGuidanceView: View {
         self.messageSecondary = messageSecondary
     }
 
+    private var trimmedTitle: String {
+        title.trimmingCharacters(in: .whitespacesAndNewlines)
+    }
+
+    private var trimmedMessage: String {
+        message.trimmingCharacters(in: .whitespacesAndNewlines)
+    }
+
+    private var trimmedSecondaryLine: String? {
+        guard let messageSecondary else { return nil }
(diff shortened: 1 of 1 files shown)


kipyin:
**Update:** Addressed the review by computing `trimmedTitle`, `trimmedMessage`, and `trimmedSecondaryLine` once at the start of `body`, deriving both the “meaningful content” check and the `Text` views from those locals so we are not re-trimming on every access. Renamed the optional binding to `secondaryLine` so it no longer shadows the trimmed string. The intent remains to hide whitespace-only onboarding cards while still showing a card when there is a real title or body (including title-only when the trimmed title is non-empty). Follow-up worth tracking: align `sectionGuidance` / `SequentialSectionPrimaryColumn` / `EditableTextSection` with the same trim-and-meaningful-content rules so the live journal path cannot drift from this component.

PR #422, opened 2026-04-14
Treat whitespace-only journal onboarding guidance as inactive

kipyin:
## Headline

Guided journal onboarding no longer pretends to show copy when the message is blank or only whitespace.

## User impact

Users see a consistent guided state: the app does not mark onboarding guidance as active when there is nothing meaningful to read. On the gratitude step, an empty or whitespace-only keyboard hint line is omitted instead of leaving an awkward blank second line if localization is missing or trimmed away.

## What changed

The onboarding presentation’s “guidance active” flag now matches what the UI would actually show: it requires a non-nil step and primary message text that is not empty after trimming leading and trailing whitespace and newlines. Section guidance uses the same trimmed-empty check for the primary message so whitespace-only strings do not produce a banner. For the optional gratitude keyboard hint, the localized secondary line is trimmed and dropped entirely when it would otherwise be blank, keeping the banner clean.

## Verification

On a Mac with Xcode and the project’s grace CLI installed, run `grace ci` (or `grace test` with the repo’s default simulator destination) to lint and build; optionally exercise the first-time journal onboarding flow in the simulator to confirm banners and hints only appear when copy is non-empty.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Tutorial/JournalOnboardingFlowEvaluator.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Tutorial/JournalOnboardingFlowEvaluator.swift
@@ -62,8 +62,10 @@ struct JournalOnboardingPresentation: Equatable {
         sectionStates: [:]
     )
 
+    /// True when guided copy is actually shown (`sectionGuidance` would be non-nil for the active step).
     var isGuidanceActive: Bool {
-        step != nil
+        guard step != nil, let message else { return false }
+        return !message.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty
     }
 
     func state(for section: JournalOnboardingSection) -> JournalOnboardingSectionState {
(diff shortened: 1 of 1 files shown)


kipyin:
Thanks for the detailed review. I added unit tests in `JournalOnboardingFlowEvaluatorTests` that lock in the whitespace rules: when the banner `message` is only whitespace or newlines, `isGuidanceActive` is false and `sectionGuidance(for:)` is nil for that step; and for the gratitude keyboard hint, `messageSecondary` is nil when the localized string is whitespace-only, and surrounding whitespace is trimmed when real copy is present. I also tightened the `isGuidanceActive` documentation so it matches behavior: guidance is considered active when trimmed body copy would actually show for the active step’s banner section (i.e., when `sectionGuidance(for: step.bannerSection)` would be non-nil), without implying other sections always get a banner. These changes keep CI honest on the new trimming logic and avoid future regressions from refactors.

PR #423, opened 2026-04-14
Align journal onboarding active state with banner-ready guidance

kipyin:
## Headline

Guided onboarding now reports “active” only when a step has real body copy, matching the banner.

## User impact

Before, `isGuidanceActive` could be true whenever a step was set, even if the message was missing or empty—so other UI could think coaching was on while no banner copy was ready. That mismatch could confuse layouts, accessibility, or logic that keys off “guidance is showing.” After this change, “active” tracks the same requirements as showing guidance: a step plus non-empty message.

## What changed

The presentation model now derives a single “guided step with message” concept: both the step and non-empty body text must be present, matching what `sectionGuidance` needs for a visible banner.

`isGuidanceActive` uses that shared rule instead of only checking for a non-nil step, and `sectionGuidance` reuses it so the active flag and per-section banner content stay aligned. A brief doc comment explains why both pieces are required.

## Verification

On a Mac with Xcode and the iOS Simulator, run `grace ci` (or `grace test` with the project’s usual destination) to lint and build. Manually spot-check Today journal onboarding: with the first guided steps, confirm guidance-related UI still activates with the expected banner copy, and that edge cases with empty or missing message (if constructible in tests or fixtures) do not mark guidance as active without visible copy.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Tutorial/JournalOnboardingFlowEvaluator.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Tutorial/JournalOnboardingFlowEvaluator.swift
@@ -62,8 +62,14 @@ struct JournalOnboardingPresentation: Equatable {
         sectionStates: [:]
     )
 
+    /// Guided step plus body copy; both are required for a visible banner (`sectionGuidance`).
+    private var guidedStepWithMessage: (step: JournalOnboardingStep, message: String)? {
+        guard let step, let message, !message.isEmpty else { return nil }
+        return (step, message)
+    }
+
     var isGuidanceActive: Bool {
-        step != nil
+        guidedStepWithMessage != nil
     }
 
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

Thanks for the thorough review. This change adds a single shared notion of an “active guided step” that requires both a step and non-empty banner copy, and wires `isGuidanceActive` to that so it can’t report guidance as on when there’s nothing to show. That aligns `isGuidanceActive` with `sectionGuidance` and closes the inconsistent UI state you called out. I agree the tests could assert the positive `isGuidanceActive` cases for gratitudes / need / person if we want stronger guardrails against future predicate drift; I can follow up in a small test-only pass if you want that locked in. CI on macOS remains the right place to run `JournalOnboardingFlowEvaluatorTests` and the rest of the suite for full confidence.

kipyin:
@cursor resolve merge conflict

PR #425, opened 2026-04-14
Harden weekly review aggregates: continuity text and rhythm history dates

kipyin:
## Headline

Weekly review insights skip bad calendar math and avoid phantom themes from empty reflection text.

## User impact

Continuity theme detection is less likely to pick up nonsense or whitespace-only signals when reading notes and reflections are both empty. If the calendar cannot compute key week boundaries safely, the rhythm history section is omitted instead of building from substitute dates that could skew the story.

## What changed

Added a small helper that trims reading notes and reflections, drops empty pieces, and joins the rest with a single space so theme extraction never receives a lone stray space. Continuity stats now feed that combined text into theme extraction instead of always inserting a space between raw fields.

In the rhythm history builder, two date steps that used fallback dates when calendar arithmetic returned nil now bail out and return no rhythm history, so we do not silently substitute week boundaries that could misrepresent the timeline.

## Verification

On a Mac with Xcode and the project CLI installed, run `grace ci` (or `grace test` with the repo’s default simulator destination) to lint, build, and exercise unit tests for the GraceNotes scheme.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Services/WeeklyReviewAggregates.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Services/WeeklyReviewAggregates.swift
@@ -146,6 +146,14 @@ private extension WeeklyReviewAggregatesBuilder {
             || !textNormalizer.trimmed(entry.reflections).isEmpty
     }
 
+    /// Joins non-empty trimmed notes and reflections without a stray lone space when both are empty.
+    func reflectionCorpusForContinuity(_ entry: Journal) -> String {
+        [entry.readingNotes, entry.reflections]
+            .map { textNormalizer.trimmed($0) }
+            .filter { !$0.isEmpty }
+            .joined(separator: " ")
+    }
+
     func buildChipStats(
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

**What changed and why**

We tightened how the weekly review builds the “reflection corpus” for continuity themes so whitespace-only notes and reflections no longer collapse into a useless single-space string—empty means empty, and real text still joins cleanly. We also updated rhythm history date stepping so when calendar arithmetic fails (only plausible at extreme representable dates), we return no rhythm history instead of silently substituting the end of the current period; the review summary already falls back to activity when rhythm history is missing. These are small, defensive fixes with no persistence or data-loss impact. Full validation is on macOS: `grace test` / `grace ci` and existing weekly review history tests; a follow-up could add a focused unit test around `reflectionCorpusForContinuity` (whitespace-only vs empty vs mixed) so a future edit can’t regress back to bad concatenation without CI noticing.

PR #426, opened 2026-04-14
Close performance traces when a startup attempt is superseded

kipyin:
## Headline

Finish superseded startup traces so Instruments and diagnostics stay honest during retries and overlapping attempts.

## User impact

If you tap Retry or a new startup begins before an older persistence attempt finishes, the stale attempt quietly returned and left its performance trace open. That skews startup timing signals and makes triage harder when something goes wrong during launch.

## What changed

Startup success and failure handlers still ignore work from an outdated attempt ID so only the current attempt updates the UI and state. When that guard fires, the code now ends the matching performance trace with an explicit superseded signal using the original trace start time, instead of returning without closing the span.

## Verification

On a Mac with Xcode, run `grace ci` (or `grace test` with the project’s usual simulator destination) to confirm the app still builds and tests pass.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Application/StartupCoordinator.swift
+++ b/GraceNotes/GraceNotes/Application/StartupCoordinator.swift
@@ -158,7 +158,10 @@ final class StartupCoordinator: ObservableObject {
         attemptID: UInt64,
         traceStart: TimeInterval
     ) {
-        guard attemptID == currentAttemptID else { return }
+        guard attemptID == currentAttemptID else {
+            PerformanceTrace.end("StartupCoordinator.startupAttempt.superseded", startedAt: traceStart)
+            return
+        }
         finishAttempt()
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

We addressed the superseded-startup path in `StartupCoordinator` so every `StartupCoordinator.startupAttempt` trace gets a matching end: when a newer attempt replaces the one in flight, we now call `PerformanceTrace.end` with the `…superseded` label before returning, instead of leaving a dangling `[BEGIN]` with no `[END]`. That keeps startup timing in logs honest and avoids inflated or misleading durations when attempts race or overlap. We did not add automated coverage for this branch in this pass; the superseded path is awkward to trigger from the public API today, and CI on macOS remains the gate for compile and `StartupCoordinatorTests`.

PR #428, opened 2026-04-14
Guard past statistics interval state updates when unchanged

kipyin:
## Headline

Advanced Settings no longer re-applies the same Past Statistics interval when storage has not actually changed.

## User impact

This avoids needless churn when the screen loads or AppStorage refreshes with values that already match the pickers, which helps keep wheel pickers and persistence from firing redundant updates. The visible setting should stay the same; the win is calmer behavior and fewer accidental side effects.

## What changed

Past statistics interval loading now computes the next quantity, unit, and All versus Custom mode from saved preferences, then exits early when that matches what is already on screen. Only when something differs does it assign into local state.

That keeps unchanged preferences from triggering unnecessary state writes that could cascade into persistence or SwiftUI change handling.

## Verification

On a Mac with the repo and Xcode, run `grace ci` (or `grace test` if you only need tests). In the app, open Settings → Advanced, choose a custom Past Statistics interval, leave the screen and return, and confirm the pickers stay put without odd resets or duplicate saves.

## Risk

Medium

## Touch class

`ui-ux`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Settings/AdvancedSettingsScreen.swift
+++ b/GraceNotes/GraceNotes/Features/Settings/AdvancedSettingsScreen.swift
@@ -164,9 +164,15 @@ struct AdvancedSettingsScreen: View {
 
     private func loadIntervalState() {
         let selection = PastStatisticsIntervalPreference.selection(fromAppStorage: intervalEncoded).validated
-        customQuantity = Self.clampedPickerQuantity(selection.quantity)
-        customUnit = selection.unit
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

No new review comments were incorporated in this pass.

## Files touched in this pass

None.

This iteration does not change any source files or project assets. There was no reviewer feedback digest to apply, so the branch is unchanged from its previous state. The existing PR description and prior commits remain the authoritative summary of the work. Any follow-up will address new review notes or scope when they arrive.

PR #429, opened 2026-04-14
Thicken settings highlight border for Differentiate Without Color

kipyin:
## Headline

Settings deep-link highlights stay visible when the system de-emphasizes color.

## User impact

People who turn on Differentiate Without Color rely on weight and contrast, not hue alone. A thin accent outline can be easy to miss. A slightly thicker border makes the highlighted settings target easier to spot without changing normal appearance for everyone else.

## What changed

The settings target highlight modifier now reads the system “Differentiate Without Color” accessibility setting. When that mode is on, the rounded highlight outline uses a heavier stroke so the selection reads clearly even when color differences are muted. When it is off, the outline matches the previous one-point stroke.

## Verification

On a Mac with Xcode, run `grace ci` (or `grace test` with the project’s default simulator destination). On device or Simulator, enable Settings → Accessibility → Display & Text Size → Differentiate Without Color, open a flow that applies the settings target highlight, and confirm the outline is visibly stronger while the feature still looks unchanged with the setting off.

## Risk

Medium

## Touch class

`ui-ux`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Settings/SettingsTargetHighlightModifier.swift
+++ b/GraceNotes/GraceNotes/Features/Settings/SettingsTargetHighlightModifier.swift
@@ -4,6 +4,7 @@ struct SettingsTargetHighlightModifier: ViewModifier {
     let isHighlighted: Bool
 
     @Environment(\.accessibilityReduceTransparency) private var reduceTransparency
+    @Environment(\.accessibilityDifferentiateWithoutColor) private var differentiateWithoutColor
 
     func body(content: Content) -> some View {
         content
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

Thanks for the deep review—agreed there are no blocking issues.

**What changed:** In `SettingsTargetHighlightModifier`, the highlight border uses a thicker stroke when *Differentiate Without Color* is on so the selection stays easy to see next to the filled highlight without relying only on color.

**Why:** That setting is meant for people who need stronger non-color cues; a slightly bolder outline makes the target clearer in Settings.

**Verification:** I’ll sanity-check Settings on a small phone layout and with *Differentiate Without Color* on and off to confirm the stroke reads well and doesn’t feel too heavy when the setting is off. I’m not adding an automated UI test for this since it’s pure styling and doesn’t change behavior contracts.

PR #430, opened 2026-04-14
Remove legacy app tour flag after one-time migration

kipyin:
## Headline

Onboarding migration now drops the old tour-seen key so settings stay tidy and predictable.

## User impact

Upgrading users still keep whether they already saw the app tour, because the value is copied to the current key first. After that one-time step, the obsolete entry is gone, which reduces leftover cruft in preferences and matches how other boolean migrations behave.

## What changed

The app tour migration used to copy the legacy “post-seed journey seen” flag into the modern “has seen app tour” key but left the old key behind. That left duplicate state in UserDefaults and was inconsistent with other one-shot boolean migrations in this area.

The migration still runs only when the new key is unset and the legacy key exists: it copies the boolean, then removes the legacy entry. Comments were updated to describe that cleanup explicitly.

## Verification

On a Mac with Xcode and the project’s grace CLI, run `grace ci` (or `grace test` with the usual simulator destination) to confirm the project still builds and tests pass.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Tutorial/JournalOnboardingProgress.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Tutorial/JournalOnboardingProgress.swift
@@ -10,7 +10,7 @@ enum JournalOnboardingStorageKeys {
     /// Legacy upgrade cohort: Today clears this after one branch resolution. Listed for iCloud continuity.
     static let legacy051GuidedBranchResolution = "journalOnboarding.pending051GuidedJournalBranchResolution"
     static let hasSeenAppTour = "journalOnboarding.hasSeenAppTour"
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

**Reply to escalated review**

We tightened `test_migrateLegacyAppTourSeenFlag_copiesTrueFromLegacyKeyOnce` so it also checks that the legacy key is absent after migration runs. That way CI would catch a regression that stopped calling `removeObject` on the old key while still copying into `hasSeenAppTour`.

For product behavior, `hasSeenAppTour` remains the single source of truth once set, so the migration stays idempotent for what users see in this build.

The downgrade scenario (an older binary that only reads the legacy flag after this build removed it) and the tiny theoretical window if the app died between `set` and `removeObject` are intentional one-way migration tradeoffs we are not optimizing for; neither affects correctness for users staying on current builds.

PR #433, opened 2026-04-14
Fix weekly review theme evidence when support text uses CJK script

kipyin:
## Headline

Supporting notes and reflections now match recurring themes correctly when only the support text is in Chinese or other CJK script.

## User impact

Weekly review theme sections can pull in evidence from reading notes and reflections. Before this change, a Latin theme label paired with CJK-only (or CJK-heavy) support text could miss valid links because Latin word-boundary rules do not apply to those scripts. Users should see fuller, more accurate theme evidence in mixed-language journals.

## What changed

The helper that decides whether a support surface semantically matches a theme used Latin-style checks whenever the theme label looked Latin, even when the support passage was CJK. That mismatch could block substring-style matches that are appropriate when word boundaries and Latin token overlap are not meaningful. The logic now treats the pair as a CJK-style match when either the theme or the support text contains Han characters, and documents why. Latin text still uses word-boundary and token overlap behavior to avoid spurious substring hits inside other words.

## Verification

On a Mac with Xcode and the iOS Simulator, run `swiftlint lint` from the repo root, then `grace ci` (or `grace test` with the project’s default destination) to confirm the GraceNotes scheme builds and tests pass.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Services/WeeklyReviewAggregatesBuilder+ThemeSections.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Services/WeeklyReviewAggregatesBuilder+ThemeSections.swift
@@ -280,7 +280,8 @@ extension WeeklyReviewAggregatesBuilder {
         let normalizedSupport = textNormalizer.normalizeThemeLabel(supportText)
         let normalizedTheme = textNormalizer.normalizeThemeLabel(themeConcept)
         guard !normalizedSupport.isEmpty, !normalizedTheme.isEmpty else { return false }
-        if containsHanCharacters(themeConcept) {
+        // CJK: word boundaries and Latin token overlap do not apply; substring match either direction.
+        if containsHanCharacters(themeConcept) || containsHanCharacters(supportText) {
             return normalizedSupport.contains(normalizedTheme) || normalizedTheme.contains(normalizedSupport)
         }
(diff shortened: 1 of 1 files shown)


kipyin:
Here is a concise PR reply you can paste:

---

**What changed and why**

We tightened `moderateSurfaceSemanticMatch` so weekly-review theme evidence still finds **Latin canonical themes** against **Han/kanji support text** (where word boundaries and token overlap often miss), without treating every mixed string as a free substring match. When the raw theme or support text includes **Han**, we still run the Latin word-boundary and token-overlap checks whenever **both normalized strings contain Latin letters**; only if those checks fail do we use bidirectional substring matching, and that substring fallback is **gated so both normalized theme and support include Han**—so a single Han character in mostly Latin text does not revive bogus Latin substring hits (for example *rest* inside *forest …*). Comments now say **Han/kanji** instead of a vague “CJK,” since detection is Unicode Han ranges (Hangul and kana-only Japanese are unchanged and still use the Latin path).

**Tests**

New cases in `WeeklyReviewAggregatesSemanticMatchTests` cover a **positive Han adjacency** match the token path would miss, **rejection** of the mixed-script *rest* / *forest 公园* false positive, and a **true positive** where Latin word boundaries still apply when both sides include Latin (*deep rest 公园*). Please run **`grace test`** (or your usual Xcode run) on a Mac to confirm the suite passes in your environment.

PR #434, opened 2026-04-14
Align completion chip accessibility label with SwiftUI locale

kipyin:
## Headline

VoiceOver now formats the sticky completion chip’s status sentence using the same locale as the journal UI.

## User impact

Screen Reader users hear section counts and wording that match the language and formatting rules of the on-screen journal, including when the UI is not using the device’s default locale. This avoids mismatches between what you see and what VoiceOver announces for the toolbar completion chip.

## What changed

The toolbar completion chip already reads locale from the SwiftUI environment for layout and measuring. The accessibility label for that chip was still building its formatted sentence with the process default locale, which could disagree with the environment in previews, tests, or any place the view hierarchy overrides locale.

Formatting now uses the same environment locale as the rest of the view. The unused UIKit import was removed because nothing in the file required it.

## Verification

On a Mac with Xcode and the project’s grace CLI installed, run `grace ci` from the repo root (or `grace test` with your usual simulator destination) to lint and build. Manually: change app or system language, open a journal day, expand the sticky completion chip label if needed, and use VoiceOver on the chip; confirm the spoken status sentence matches the visible language and number formatting.

## Risk

Medium

## Touch class

`ui-ux`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Views/JournalCompletionBarChip.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Views/JournalCompletionBarChip.swift
@@ -1,5 +1,4 @@
 import SwiftUI
-import UIKit
 
 /// Compact completion control for the navigation bar: **capsule** fill (tier colors).
 ///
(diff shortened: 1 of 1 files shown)


kipyin:
Thanks for the catch on locale alignment. In `JournalCompletionBarChip`, the section-counts accessibility string now loads `journal.share.sectionCountsSentence` with **`String(localized:…, locale:)`** using the same **`@Environment(\.locale)`** value we already pass into **`String(format:locale:…)`** for the interpolated numbers. That way, when SwiftUI’s locale in this subtree differs from the system default, the sentence template and the formatted counts stay in the same language and numbering rules instead of mixing `.current` for the format string with the environment for the values.

PR #436, opened 2026-04-14
Harden demo seeding: rollback failed saves and align day lookup

kipyin:
## Headline

Demo builds now update the same journal row the app considers canonical for a day, and failed saves leave no partial inserts behind.

## User impact

In the Demo scheme, seeding is less likely to leave the database half-written if saving fails, which avoids confusing or inconsistent sample data on the next launch. When duplicate journal rows exist for one calendar day, demo updates now match what Past and Today show instead of whichever row happened to sort first.

## What changed

When persisting demo seed data fails, the change rolls back the model context so unsaved inserts and edits are not left lingering in memory. The helper that finds an existing journal for a calendar day no longer uses a one-off fetch; it calls the same repository API the product uses to resolve a day’s journal, so upserts follow the app’s canonical choice if multiple rows share a day. Documentation comments were updated to describe that shared behavior.

## Verification

On a Mac with Xcode, build and run the GraceNotes (Demo) scheme, relaunch to confirm demo data seeds as expected, and run `grace ci` (or `grace test` with the project’s default simulator destination) to match CI.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Data/Persistence/DemoDataSeeder.swift
+++ b/GraceNotes/GraceNotes/Data/Persistence/DemoDataSeeder.swift
@@ -31,6 +31,7 @@ enum DemoDataSeeder {
             UserDefaults.standard.set(seedVersion, forKey: seedVersionKey)
             PerformanceTrace.end("DemoDataSeeder.seedIfNeeded", startedAt: seedTrace)
         } catch {
+            context.rollback()
             PerformanceTrace.end("DemoDataSeeder.seedIfNeeded.failed", startedAt: seedTrace)
             assertionFailure("Failed to seed demo database: \(error)")
         }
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

**What changed and why**

We now roll back the SwiftData context when `save()` fails on the demo-only seeding path, so a failed persist does not leave the in-memory store half-seeded. Demo day lookup is delegated to `JournalRepository.fetchEntry(dayStart:context:)`, so demo upserts use the same canonical duplicate selection as Past/Today instead of the old `createdAt`-ordered fetch that could touch a different duplicate row. SwiftData save/rollback is not covered by Linux CI; if we need extra confidence, a short smoke run on macOS with the **GraceNotes (Demo)** scheme (including an optional forced save failure) is the right check. The debug-only `assertionFailure` path on fetch failure that can look like “missing journal” is unchanged from before this diff.

PR #437, opened 2026-04-14
Reset completion info card state without animating the hidden step

kipyin:
## Headline

The journal completion card’s intro animation stays crisp and predictable instead of fighting inherited motion.

## User impact

When the completion info card plays its first-time reveal, the transition should feel like one smooth motion into place. Previously, resetting visibility could pick up implicit animation and make the intro feel wrong or janky. The card still animates in on purpose; only the reset to the hidden state is immediate.

## What changed

The entry animation for the completion info card sets visibility to hidden immediately before running the explicit ease-out animation back to visible. That reset is now wrapped in a SwiftUI transaction that turns animations off for that step alone, so nothing else accidentally animates the hide. The following reveal still uses the same timed animation as before.

## Verification

On a Mac with the repo’s dev setup, run `grace ci` (or `grace test` with your usual simulator destination) to confirm lint and the GraceNotes scheme still build and tests pass. In the simulator, open a flow that shows the completion info card with motion enabled and confirm the card’s text fades and moves in once without a stray or doubled transition.

## Risk

Medium

## Touch class

`ui-ux`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Views/CompletionInfoCard.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Views/CompletionInfoCard.swift
@@ -83,7 +83,11 @@ struct CompletionInfoCard: View {
         }
         didPlayEntryAnimation = true
 
-        contentVisible = false
+        var transaction = Transaction()
+        transaction.disablesAnimations = true
+        withTransaction(transaction) {
+            contentVisible = false
+        }
 
         withAnimation(.easeOut(duration: 0.24)) {
             contentVisible = true


kipyin:
**Sentry — review feedback**

Thanks for the thorough pass—glad there’s nothing blocking.

We adjusted the completion card so its initial “hidden” state is applied **without** inheriting the surrounding insert/transition animation, then used the existing explicit animation only for the reveal. That keeps asymmetric inserts from accidentally animating opacity/offset on the invisible setup step, while the user-visible entrance still runs as intended.

I’ll do a quick manual check on **iPhone Simulator** for the morph/matched-geometry case if `CompletionInfoCard` is ever shown with `showMorph: true`, and I’ll sanity-check rapid dismiss during the short entrance so nothing odd happens if the view tears down mid-animation. No data or persistence paths were touched here—this is animation behavior only—so I’m not adding automated coverage for that micro-interaction in this change.

PR #438, opened 2026-04-14
Harden journal keyboard geometry and first-responder caret handling

kipyin:
## Headline

More reliable keyboard overlap and caret tracking while journaling so the editor scrolls correctly.

## User impact

The Today journal screen should keep the cursor and content aligned with the keyboard more consistently, including when the system reports keyboard frames in the documented NSValue form. When focus lands on an internal text subview, caret-based logic can still find the real text field, so you get fewer unnecessary full-section scrolls or missed caret nudges while typing.

## What changed

Keyboard overlap height now reads the end keyboard frame from notifications in a way that matches UIKit’s documented userInfo shape, instead of relying on a direct CGRect cast that often fails. Caret visibility and inner text view scrolling now resolve the owning UITextView or UITextField by walking up from whichever view is actually first responder, which covers cases where UIKit reports an internal subview as first responder. Together these changes make keyboard-related inset and scroll decisions reflect real geometry and caret position more often.

## Verification

On a Mac with Xcode and the iOS Simulator, run `grace ci` (or `grace test` with the project’s default destination) and manually smoke-test the Today journal: show and dismiss the keyboard, type in multiline notes and single-line fields, and confirm the caret stays visible without jarring extra scrolling.

## Risk

Medium

## Touch class

`ui-ux`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Views/JournalKeyboardScrollSupport.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Views/JournalKeyboardScrollSupport.swift
@@ -78,7 +78,7 @@ enum JournalKeyboardScrollMetrics {
 /// Keyboard overlap with the key window; sizes extra scroll affordance without guessing global safe-area behavior.
 enum JournalKeyboardOverlapReader {
     static func overlapHeight(from notification: Notification) -> CGFloat {
-        guard let frame = notification.userInfo?[UIResponder.keyboardFrameEndUserInfoKey] as? CGRect else {
+        guard let frame = keyboardFrameEnd(from: notification) else {
             return 0
         }
         guard let window = JournalKeyWindowReader.keyWindow() else {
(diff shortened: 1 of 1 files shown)


kipyin:
Addressed the review in two ways: keyboard notifications now parse `keyboardFrameEndUserInfoKey` through the documented `CGRect` and `NSValue` paths; if the value is present but has an unexpected type, we log a warning and in Debug assert so silent “zero overlap” failures don’t go unnoticed. For caret geometry, the superview walk only treats an ancestor `UITextView` or `UITextField` as the owning control when the first responder is a descendant and its origin maps inside that view’s bounds (with the usual direct-cast paths unchanged), so we don’t read caret rects from the wrong control in odd nested hierarchies. Added brief doc comments on those helpers describing the intended UIKit/SwiftUI invariants; automated tests still don’t cover real keyboard internals, so validation remains manual on device/simulator with `grace test` / `grace ci` on a Mac.

PR #439, opened 2026-04-14
Scale inline journal editing layout for Dynamic Type

kipyin:
## Headline

Inline sentence editing spacing now grows with the user’s text size instead of staying fixed.

## User impact

People who use larger or smaller text in Settings get editor insets, vertical offsets, bottom spacing, and the dismiss tap area scaled so the layout stays proportionate and easier to use. Ambient dimming of unfocused rows is unchanged.

## What changed

The shared `SequentialSectionInlineLayout` values that control the morphing editor’s horizontal inset, vertical offset, bottom spacing, and minimum height for the bottom tap catcher were switched from fixed constants to values scaled with `UIFontMetrics` for the body text style (replacing the prior `Foundation` import with `UIKit`).

Opacity for non-focused content during inline editing stays a fixed value and was not altered.

## Verification

On a Mac with Xcode and the project’s `grace` CLI installed, run `grace ci` (or `grace test` with the usual simulator destination from `gracenom-dev.toml`) to confirm the app builds and automated checks pass. On device or Simulator, open journal inline editing with different Dynamic Type sizes in Display & Text Size and confirm spacing and dismiss behavior still feel right.

## Risk

Medium

## Touch class

`ui-ux`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Views/SequentialSectionInlineLayout.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Views/SequentialSectionInlineLayout.swift
@@ -1,12 +1,26 @@
-import Foundation
+import UIKit
 
 /// Shared metrics for inline sentence editing (entry-row editor and add morph composer).
 enum SequentialSectionInlineLayout {
-    static let editorMorphHorizontalInset: CGFloat = 6
-    static let editorMorphVerticalOffset: CGFloat = 8
-    static let editorBottomSpacing: CGFloat = 10
+    private static let bodyMetrics = UIFontMetrics(forTextStyle: .body)
+
+    static var editorMorphHorizontalInset: CGFloat {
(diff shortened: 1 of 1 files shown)


kipyin:
**Response to review**

Spacing for the inline editor and add-morph row now uses **body-scaled** metrics on purpose: `InlineSentenceEditorField` drives its `UITextView` with the same **body** Dynamic Type curve (`bodyUIFont()`), so horizontal inset, vertical offset, and bottom spacing track that typography instead of fixed points. The bottom tap catcher still grows with Dynamic Type for a reliable dismiss target, but its minimum height is **capped** using a fraction of the screen height so it does not turn into an enormous empty region at the largest accessibility sizes. **Verification:** `swiftlint lint` on `SequentialSectionInlineLayout.swift` is clean; please sanity-check on device or Simulator with several Dynamic Type sizes (including the largest) for row alignment, scroll feel, and tap-to-dismiss.

PR #440, opened 2026-04-14
Harden weekly review narrative summary against empty observations

kipyin:
## Headline

Weekly recap copy now falls back when an insight observation is only whitespace.

## User impact

The weekly review narrative is less likely to show a blank or meaningless line when a second insight is empty or localization yields only spaces. When two distinct themes are summarized together, the combined line is only shown if it has real text after trimming.

## What changed

The builder used to return raw second-insight text in several two-insight paths, so whitespace-only or empty strings could still become the visible summary. A small helper now trims whitespace and treats an empty result as missing, so the summary prefers the other insight or skips empty combined strings instead of surfacing junk. The single-insight path uses the same trimming rule for consistency.

## Verification

On a Mac with Xcode, run `grace ci` (or `grace test`) from the repo root to lint and exercise the GraceNotes scheme in the simulator.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Services/WeeklyInsightCandidateBuilder.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Services/WeeklyInsightCandidateBuilder.swift
@@ -101,7 +101,7 @@ struct WeeklyInsightCandidateBuilder {
             let first = insights[0].observation
             let second = insights[1].observation
             if !observationsAreEffectivelyDuplicate(first, second) {
-                return second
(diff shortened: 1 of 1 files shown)


kipyin:
We tightened the two-insight path where both insights have different non-empty primary themes: after building the combined localized line we trim it and, if that string is empty, fall back to the first insight’s observation and then the second’s so the narrative does not go missing while observation text still appears elsewhere. We added unit tests for two non-duplicate observations when the second observation is empty or only whitespace, asserting we return the first insight’s trimmed line. Together this addresses the review’s consistency note and closes the testing gaps called out for those branches.

PR #442, opened 2026-04-14
Align theme language gate with analysis prefix and bound work

kipyin:
## Headline

Past tab theme chip language detection now measures text the same way the analyzer does, without scanning huge journals.

## User impact

Very long journals no longer trigger expensive full-corpus counting just to decide if there is enough text to pick English or Chinese theme labels. The “enough meaningful text” rule now matches the portion of the journal that language detection actually uses, so chip language stays consistent with the model and avoids misleading picks from ignored tail content.

## What changed

The helper that checks whether the journal has enough non-whitespace text before running language detection was updated to count only within the same grapheme-limited prefix used for analysis. That caps work on pathological or enormous corpora and keeps the minimum-length gate aligned with the sample passed to the language recognizer. A zero or negative minimum is treated as “always enough,” avoiding pointless iteration when the threshold is disabled.

## Verification

On a Mac with Xcode and the project’s grace CLI installed, run `grace ci` (or `grace test` with the repo’s default simulator destination) to confirm the GraceNotes scheme still builds and tests pass.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Services/ReviewJournalThemeLanguageResolver.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Services/ReviewJournalThemeLanguageResolver.swift
@@ -53,18 +53,26 @@ struct ReviewJournalThemeLanguageResolver: ReviewJournalThemeLanguageResolving {
     /// Upper bound on text fed to `NLLanguageRecognizer` and script tie-break (prefix by extended grapheme cluster).
     private static let maximumAnalysisGraphemes = 50_000
 
+    /// Counts non-whitespace graphemes only in the same prefix we analyze, so the threshold matches
+    /// `normalizedAnalysisSample` and pathological corpora cannot spend unbounded time here.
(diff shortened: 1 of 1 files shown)


kipyin:
**What changed and why**

We aligned the “enough meaningful text” gate with the same first 50,000 extended grapheme clusters we feed to `NLLanguageRecognizer` and the script tie-break, so the threshold matches what we actually analyze and `hasEnoughMeaningfulGraphemes` cannot scan unbounded input on pathological journals. The initializer still clamps `minimumMeaningfulGraphemes` with `max(0, …)`, and when the minimum is zero we skip the gate. File-level comments now spell out that the minimum applies to the **analysis prefix**, not the whole trimmed corpus, including the tradeoff that a long sparse head with dense language-specific text only after that prefix can still fall back to English. We added a focused unit test for that sparse-head / dense-tail case so the prefix-only behavior stays locked in. Full verification is on macOS with `grace test` / `grace ci` (Linux CI does not run `GraceNotesTests`).

PR #443, opened 2026-04-14
Disable empty section-distribution segments and align VoiceOver

kipyin:
## Headline

Empty section segments no longer behave like buttons or are announced as buttons in Review history.

## User impact

In Review history insights, section distribution segments with zero lines could still feel tappable and VoiceOver could announce them as buttons, which was misleading and could open a drilldown with nothing to show. Those segments are now inactive when the count is zero, and accessibility traits match the empty-state hints.

## What changed

The history section distribution strip uses buttons for Gratitudes, Needs, and People so people can drill into each kind of content. Segments with a line count of zero were still enabled as buttons, which was inconsistent with the growth stages skyline (which already disables zero-count columns) and weakened trust for VoiceOver users. The change disables interaction when a segment’s count is zero and drops the button trait in that case so assistive technology does not present the segment as an actionable control. The same behavior is applied in both strip layout variants that render those segments.

## Verification

On a Mac with Xcode and the iOS Simulator, run `grace ci` from the repo root (or `grace test` with the default destination in `gracenotes-dev.toml`). Manually open Review → History insights with data where one section count is zero; confirm the empty segment does not respond to taps and, with VoiceOver, that it is not announced as a button while the empty hint still reads sensibly.

## Risk

Medium

## Touch class

`ui-ux`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Views/ReviewHistoryInsightsPanels.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Views/ReviewHistoryInsightsPanels.swift
@@ -410,6 +410,7 @@ private struct ReviewHistorySectionStrip: View {
                             .frame(width: segmentWidths[index], height: stripHeight)
                         }
                         .buttonStyle(PastTappablePressStyle())
+                        .disabled(item.count == 0)
                         .accessibilityLabel(
                             sectionMixAccessibilityLabel(
                                 kind: item.kind,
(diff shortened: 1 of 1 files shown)


kipyin:
Addressed the VoiceOver issue for zero-count trend segments. Disabled `Button`s still advertise the default button trait, and `accessibilityAddTraits([])` does not remove it, so empty segments could still appear as buttons in the rotor. We now call `accessibilityRemoveTraits(.isButton)` when `item.count == 0` on both the strip and legend paths, so VoiceOver treats those segments as non-interactive text instead of buttons. This mirrors the pattern used elsewhere in the file where only actionable controls get the button trait. I have not run `grace test` / `grace ci` here; please run them on a Mac and spot-check VoiceOver so empty segments no longer show under “Buttons.”

PR #444, opened 2026-04-14
Hop StartupCoordinator work to MainActor from async startup tasks

kipyin:
## Headline

Startup loading and reassurance updates now apply on the main thread after background work, with clearer handling when an attempt is cancelled.

## User impact

Startup should feel the same, but the app is less likely to hit subtle threading or Swift concurrency issues while persistence loads and the loading messages rotate. If you retry or the coordinator is torn down mid-startup, traces and state should line up more reliably instead of racing the UI.

## What changed

Background `Task` closures no longer capture `self` up front for the whole startup, copy rotation, and reassurance timers. Instead, the persistence factory and timing intervals are captured as plain values, async work runs off the actor, and success, failure, copy advances, and reassurance transitions are dispatched with `MainActor.run` so `@MainActor` state and `@Published` updates stay on the main actor. When the coordinator is gone before that hop (for example, a superseded attempt), performance traces are ended with a superseded event instead of dropping the trace.

## Verification

On a Mac with Xcode, run `grace ci` (or `grace test` with the project’s default simulator destination) to lint, build, and exercise tests; smoke the app launch and a retry after a simulated startup failure if you want extra confidence.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Application/StartupCoordinator.swift
+++ b/GraceNotes/GraceNotes/Application/StartupCoordinator.swift
@@ -82,52 +82,68 @@ final class StartupCoordinator: ObservableObject {
 
         currentAttemptID += 1
         let attemptID = currentAttemptID
+        let runPersistence = persistenceFactory
+        let copyRotationInterval = timing.copyRotationInterval
+        let reassuranceDelay = timing.reassuranceDelay
         isStartingUp = true
         loadingCopyIndex = 0
         reassuranceCopyIndex = 0
         phase = .loading
         startupMessage = loadingMessage(for: loadingCopyIndex)
-        startCopyRotation(for: attemptID)
-        scheduleReassuranceTransition(for: attemptID)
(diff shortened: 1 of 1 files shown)


kipyin:
Thanks for the review. When persistence finishes after the coordinator has been deallocated, startup traces now end as `StartupCoordinator.startupAttempt.teardown` instead of reusing the “superseded” outcome, so metrics and ops logs are not confused with a newer attempt taking over. For retryable startup failures, the UI always shows the fixed `startup.error.setupFailed` copy while the full error is written to the unified logger for debugging, avoiding `LocalizedError` descriptions that could expose file paths or other internal detail. I did not change the “superseded” trace when `attemptID` no longer matches the current attempt—that still reflects a competing attempt finishing late. Validation on this branch is through `grace test` / `grace ci` on macOS with the project’s default simulator destinations.

PR #445, opened 2026-04-14
Trim whitespace when decoding journal completion level from legacy strings

kipyin:
## Headline

Journal growth-stage labels from older data now decode reliably instead of falling back to Soil.

## User impact

If a saved completion label had leading or trailing spaces or line breaks, the app could misread it and treat the day as Soil (empty) even when it was not. That could skew history, streaks, or any UI that depends on the stored level. After this change, those values normalize the same way users expect when reading text.

## What changed

The legacy string decoder for journal completion level now trims whitespace and newlines before matching synonyms and lowercasing. That makes persisted labels from older app versions or odd inputs align with the known Soil → Bloom names instead of silently defaulting to Soil. No other journal model behavior was changed.

## Verification

On a Mac with Xcode, run `grace ci` (or `grace test`) from the repo root to lint and build against the project’s simulator destinations.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Data/Models/Journal.swift
+++ b/GraceNotes/GraceNotes/Data/Models/Journal.swift
@@ -25,7 +25,8 @@ extension JournalCompletionLevel: Codable {
 
     /// Maps persisted raw strings from older app versions and unknown values.
     init(decodingLegacyRawValue raw: String) {
-        switch raw.lowercased() {
+        let normalized = raw.trimmingCharacters(in: .whitespacesAndNewlines).lowercased()
+        switch normalized {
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

## Summary

This change hardens decoding of persisted `JournalCompletionLevel` values when loading older journal data. Before matching against known legacy strings, the raw value is trimmed of leading and trailing whitespace and newlines, then lowercased. That avoids misclassifying completion level when stored strings include stray spaces or line breaks, so completion state stays stable across upgrades and edge-case data.

PR #446, opened 2026-04-14
Fix spurious weekly review theme matches in mixed-language entries

kipyin:
## Headline

Weekly review theme evidence no longer treats Latin words as matching inside other words when notes also include Chinese or Japanese characters.

## User impact

Users with bilingual journaling may have seen unrelated reading notes or reflections attached to a theme because a short English theme could match as a substring (for example, “rest” inside “forest”). That weakens trust in recurring themes and trend evidence. The review should now only use loose substring matching when the theme itself needs CJK-style handling, while English themes stay on stricter word-based checks.

## What changed

Supporting evidence matching previously took a “substring either direction” shortcut whenever the support text contained CJK characters, even if the distilled theme label was Latin-only. That path skipped the Latin safeguards that avoid in-word hits. The logic now reserves substring matching for themes that actually contain Han characters, and routes Latin-only themes through the existing Latin-safe path (word-boundary checks, then multi-word token overlap) regardless of whether the passage mixes scripts. The Latin matching body was extracted into a small helper to keep the behavior obvious and reusable.

## Verification

On a Mac with the repo’s toolchain, run `grace ci` (or `grace test` with the project’s default simulator destination) to lint and build; spot-check a weekly review where reflections or reading notes mix English with Chinese or Japanese and confirm Latin themes do not pick up incidental substrings.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Services/WeeklyReviewAggregatesBuilder+ThemeSections.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Services/WeeklyReviewAggregatesBuilder+ThemeSections.swift
@@ -280,11 +280,20 @@ extension WeeklyReviewAggregatesBuilder {
         let normalizedSupport = textNormalizer.normalizeThemeLabel(supportText)
         let normalizedTheme = textNormalizer.normalizeThemeLabel(themeConcept)
         guard !normalizedSupport.isEmpty, !normalizedTheme.isEmpty else { return false }
-        // CJK: word boundaries and Latin token overlap do not apply; substring match either direction.
-        if containsHanCharacters(themeConcept) || containsHanCharacters(supportText) {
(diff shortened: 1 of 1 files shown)


kipyin:
Added a regression test in `WeeklyReviewAggregatesMostRecurringTests` that locks in the mixed-script behavior reviewers asked for: a Latin-only recurring theme (“Rest”) must not gain supporting evidence from reading notes or reflections where the only overlap is a substring inside another English word (“rest” inside “forest”), even when that surface also includes Han characters. The same test includes a positive control so we still expect supporting evidence when the Latin match is a real whole-word occurrence alongside CJK (e.g. “I need rest 休息”). This directly covers the false-positive path that previously relied on bidirectional substring matching whenever support text contained Han. I have not run `grace test` here; please run it on macOS to confirm the suite passes in your environment.

PR #447, opened 2026-04-14
Fix Past search match IDs for whole-field journal text

kipyin:
## Headline

Past search results now stay correctly identified when reading notes or reflections change.

## User impact

Whole-field matches (for example reading notes and reflections) used to share one identity per day and section, so SwiftUI could treat different text as the same row and reuse or update the wrong result. Identities now reflect the actual text, so lists and updates behave predictably after edits.

## What changed

Search match rows for whole-field results no longer use only the journal entry and section to form their identifier. The app now appends a short fingerprint derived from the field’s text so two different strings under the same entry and section are distinct. Line-based matches that already tie to a specific `Entry` item are unchanged. CryptoKit is used only to compute that compact hash; the full note text is not embedded in the id string.

## Verification

On a Mac with Xcode, run `grace ci` (or `grace test`) per the repo’s default simulator destinations. Manually, open Past search, edit a reading note or reflection, search again, and confirm results list and selection behave correctly without stale or duplicated rows.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Data/JournalSearchMatch.swift
+++ b/GraceNotes/GraceNotes/Data/JournalSearchMatch.swift
@@ -1,3 +1,4 @@
+import CryptoKit
 import Foundation
 
 /// One searchable line or note row tied to a journal day, for Past search results.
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

Thanks for flagging the gap. I added a `JournalRepositoryTests` case for **reading notes** (or **reflections**, depending on what landed in the branch) that mirrors the gratitudes chip test: it seeds a journal with a fixed entry id and non-empty field text, runs `searchMatches` twice, and asserts the match `id` is stable and matches the `journalEntryId|source|fingerprint` pattern (16 hex characters from the content hash). That locks in the `fieldMatchId` contract for whole-field rows so CI will catch accidental changes to hashing, length, or source tagging. I validated locally with `grace test` on the Mac simulator destination used in this repo.

PR #448, opened 2026-04-14
Fix prior-week range for review insights across DST boundaries

kipyin:
## Headline

Review insights now pick the true previous calendar week when daylight saving time shifts week length.

## User impact

The “previous week” comparison for review insights could be off by a day around DST changes because subtracting a fixed week in seconds does not always equal seven local days. Users comparing this week to last week get a consistent seven-day window aligned to local midnight and the calendar.

## What changed

The helper that computes the period immediately before the current review week no longer relies on subtracting seven days from the week start or on a fixed 604,800-second offset. It anchors on the instant just before the current week begins, takes that day’s start-of-day, then steps back six calendar days so the range is exactly seven local days ending where the current week starts. A short comment explains why fixed-second math is unsafe across offset changes.

## Verification

On a Mac with Xcode and the project’s grace CLI installed, run `grace ci` from the repo root (or `grace test` with the default simulator destination from gracenotes-dev.toml). Optionally spot-check review insights around a DST transition week in the simulator to confirm “previous week” matches expectations.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Services/ReviewInsightsPeriod.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Services/ReviewInsightsPeriod.swift
@@ -24,8 +24,12 @@ enum ReviewInsightsPeriod {
     /// The calendar week immediately before `current` (the seven local days whose end abuts
     /// `current.lowerBound`).
     static func previousPeriod(before current: Range<Date>, calendar: Calendar) -> Range<Date> {
-        let previousStart = calendar.date(byAdding: .day, value: -7, to: current.lowerBound)
-            ?? current.lowerBound.addingTimeInterval(-604_800)
+        // Anchor on the last instant before the current week so `startOfDay` + day offsets match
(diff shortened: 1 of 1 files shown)


kipyin:
**Update:** `previousPeriod(before:)` no longer uses a fixed 604,800-second stride, which could misalign “prior week” across daylight saving changes. It now derives the prior window from the last local day before the current period’s lower bound and steps back six calendar days so the range is exactly seven local days ending at `current.lowerBound`, with a rolling seven-day fallback if calendar math returns nil (so we never collapse to a one-day range). Added a unit test using **America/New_York** with a reference time after the March 2026 spring-forward transition, asserting `previous.upperBound == current.lowerBound`, start-of-day alignment, and a seven-day span—so future refactors can’t reintroduce GMT-only or fixed-second week math without failing tests. I validated this with `grace test` on macOS per repo guidance; Linux agents can’t run the Xcode test bundle.

PR #451, opened 2026-04-14
Harden journal tutorial UserDefaults migration for plist/sync booleans

kipyin:
## Headline

Tutorial hint flags migrate reliably when defaults store booleans as numbers or odd types.

## User impact

Fewer cases where onboarding hints or celebration flags come back after an update or sync, or get stuck in a bad state. The app keeps your tutorial progress consistent when old keys are migrated to the newer names.

## What changed

The legacy-key migration path no longer relies only on raw presence checks plus generic boolean reads. It now reads stored values as real booleans when they are Bool or NSNumber (common with plists and some sync paths), and only copies from a legacy key into the new key when the new key does not already hold a well-formed boolean. That avoids mis-reading corrupted or unexpected types, which could previously drop the legacy value without writing a correct new value.

## Verification

On a Mac with Xcode, run `grace ci` (or `grace test` with the project’s default simulator destination) to confirm the app builds and tests pass; optionally exercise a build that runs `migrateLegacyKeysIfNeeded` after seeding UserDefaults with legacy keys stored as 0/1 or Bool to confirm hints stay dismissed as expected.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Tutorial/JournalTutorialProgress.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Tutorial/JournalTutorialProgress.swift
@@ -36,10 +36,27 @@ enum JournalTutorialStorageKeys {
         }
     }
 
+    /// Interprets stored booleans whether `UserDefaults` returns `Bool` or `NSNumber` (plist / sync).
+    private static func optionalBool(forKey key: String, in defaults: UserDefaults) -> Bool? {
+        guard let object = defaults.object(forKey: key) else { return nil }
+        switch object {
+        case let value as Bool:
+            return value
+        case let number as NSNumber:
+            return number.boolValue
+        default:
+            return nil
+        }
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

_No reviewer notes were addressed in this update._

## Files touched in this pass

_None—this is a summary comment for the branch as it stands._

This branch tightens weekly review behavior around **daylight saving time**, so the “prior week” window for review insights stays correct when local midnight shifts. It also **routes the completion badge info card’s tint through `JournalCompletionLevel`**, so the card’s color matches the user’s completion state consistently. **Journal tutorial progress** logic was refined for clearer, more reliable behavior during the guided flow. **Search** and **weekly review aggregation** paths were adjusted in support of those fixes. Together, these changes reduce edge-case bugs in insights and make completion and tutorial UI line up with the underlying data.

PR #452, opened 2026-04-14
Reuse prepared haptic generators for Past tap button styles

kipyin:
## Headline

Past journal taps use shared, prepared haptic feedback instead of creating new generators on every press.

## User impact

Tapping items in Past and related sheets should feel the same or slightly more consistent, with less work on each tap. This mainly improves responsiveness and reliability by avoiding repeated allocations and by preparing the Taptic Engine before each impact, which can matter when scrolling and tapping quickly.

## What changed

The Past tappable press and toolbar Done button styles previously created a new impact feedback generator every time you pressed a button. The change introduces a small private helper that keeps one medium and one light generator and calls prepare before each impact, matching UIKit guidance for predictable latency. Visual press animations and Reduce Motion behavior are unchanged; only how haptics are produced under the hood was tightened.

## Verification

On a Mac with Xcode, run `grace ci` (or `grace test` if that matches your usual check) from the repo root to confirm lint and simulator build pass. Manually open Past, tap list items and the Done-style toolbar control, and confirm haptics still fire when Reduce Motion is off and stay suppressed when Reduce Motion is on.

## Risk

Medium

## Touch class

`ui-ux`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Views/PastTappablePressStyle.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Views/PastTappablePressStyle.swift
@@ -1,6 +1,22 @@
 import SwiftUI
 import UIKit
 
+/// Shared generators so taps avoid per-press allocation; `prepare()` keeps latency predictable (UIKit guidance).
+private enum PastTappableHaptics {
+    private static let mediumImpact = UIImpactFeedbackGenerator(style: .medium)
+    private static let lightImpact = UIImpactFeedbackGenerator(style: .light)
+
+    static func playMediumImpact() {
+        mediumImpact.prepare()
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

Thanks for the deep review. This updates the button haptics so we keep two shared `UIImpactFeedbackGenerator` instances instead of creating a new one on every press, call `prepare()` right before `impactOccurred()` to follow UIKit’s guidance, and leave the same safeguards in place (`reduceMotion` and firing only on the press leading edge). That should make the feedback path a bit cheaper and more predictable without changing when haptics run. You’re right that haptics still need to be checked on a real device; CI won’t catch feel. If we ever need haptics scoped per window or scene, we can revisit long-lived generators, but that’s not required for this change.

PR #453, opened 2026-04-14
Canonicalize peek viewport height to avoid negative zero in layout

kipyin:
## Headline

Review History calendar peek heights now normalize edge-case floating-point values so layout stays predictable.

## User impact

Rare floating-point edge cases could leave the calendar peek with a negative-zero height signal, which can confuse layout or downstream comparisons. Normalizing the clamped height keeps the feathered calendar grid sizing stable and consistent for reviewers browsing history.

## What changed

The peek viewport height helper still rejects non-finite values and never exceeds the remaining space, but after clamping to a non-negative value it now canonicalizes the result so IEEE negative zero cannot propagate into frame sizing for the calendar grid. Documentation was updated to explain why that normalization matters for layout.

## Verification

On a Mac with the project toolchain, run `grace ci` (or `grace test` with the repo’s default simulator destination) to confirm the GraceNotes scheme still builds and tests pass.

## Risk

Medium

## Touch class

`ui-ux`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Views/ReviewHistoryDrilldownPeekMetrics.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Views/ReviewHistoryDrilldownPeekMetrics.swift
@@ -16,9 +16,10 @@ enum ReviewHistoryDrilldownPeekMetrics {
     /// Uses at most the non-negative space remaining; never inflates above `remainingHeight`.
     ///
     /// Non-finite values (NaN or infinity) collapse to `0` so invalid geometry math cannot propagate into
-    /// frame sizes for the calendar grid.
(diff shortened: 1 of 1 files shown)


kipyin:
Added a focused regression test for `clampedViewportHeight` that feeds IEEE negative zero (`CGFloat(signOf: -1, magnitudeOf: 0)`) and checks the result matches canonical positive zero via `bitPattern`, so a future refactor cannot drop the `abs` (or equivalent) without CI failing—`XCTAssertEqual` to `0` alone would not catch that. This directly addresses the review feedback on locking canonical `+0.0` for the drill-down peek height helper. **Verification:** `grace test` (or the project’s usual `PastDrilldownCalendarLayoutTests` run on your Mac).

PR #454, opened 2026-04-14
Stabilize journal export ordering and export filenames

kipyin:
## Headline

Journal data exports now break ties with native UUID order and stamp filenames with a deterministic UTC timestamp.

## User impact

When two entries share the same day and creation time, the export order stays aligned with Swift’s UUID ordering instead of string-based UUID text. Export filenames keep the same human-readable pattern while avoiding locale-dependent date formatting quirks.

## What changed

The export archive’s secondary sort now compares journal IDs directly instead of comparing their string forms, which is a clearer tie-breaker and matches typical UUID ordering.

The timestamp embedded in export filenames is still Gregorian and UTC, but it is assembled from calendar components with fixed-width numeric formatting instead of a DateFormatter. That keeps the filename stamp predictable and less sensitive to formatting or locale edge cases.

## Verification

On a Mac with Xcode, run `grace ci` (or `grace test` per project defaults). Optionally trigger a journal data export from Settings and confirm the generated JSON filename uses the expected `yyyyMMdd-HHmmss` UTC-style segment before the random UUID suffix.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Settings/Services/JournalDataExportService.swift
+++ b/GraceNotes/GraceNotes/Features/Settings/Services/JournalDataExportService.swift
@@ -39,7 +39,7 @@ struct JournalDataExportService {
             if lhs.createdAt != rhs.createdAt {
                 return lhs.createdAt < rhs.createdAt
             }
-            return lhs.id.uuidString < rhs.id.uuidString
+            return lhs.id < rhs.id
         }
         return JournalDataExportArchive(
             schemaVersion: JournalDataExportArchive.currentSchemaVersion,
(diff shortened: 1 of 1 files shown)


kipyin:
## Reply to review

We documented that when `entryDate` and `createdAt` match, the tie-break uses `UUID`’s native `Comparable` order (RFC 4122 bytes), which is stable in-app but can differ from older exports that compared `uuidString`; import semantics are unchanged. Export filenames now build the `yyyyMMdd-HHmmss` segment with one shared `DateFormatter` (Gregorian calendar, UTC, `en_US_POSIX`) so stamps stay locale-insensitive and we avoid `?? 0` on date components that could quietly yield bogus zero segments. Added tests that lock in tie-break order against `UUID` sorting and assert known UTC filename stamps for fixed `Date` values. GMT uses `TimeZone(secondsFromGMT: 0)` with a fallback so we don’t rely on a force-unwrap for a valid time zone.

PR #456, opened 2026-04-14
Harden stored reminder time parsing from UserDefaults

kipyin:
## Headline

Daily reminders now pick up your saved time even when UserDefaults stores it as a bridged number.

## User impact

If the app’s stored reminder clock time was bridged from UserDefaults as NSNumber, the old cast to TimeInterval could fail and silently fall back to the default time, so your notification might not match the time you chose. This path now reads the value in the forms the system actually persists and ignores corrupt or non-finite numbers, so rescheduling stays aligned with your reminder settings.

## What changed

The routine that refreshes the pending repeating reminder’s notification body no longer assumes UserDefaults returns a value you can cast directly to TimeInterval. A dedicated helper loads the key’s raw object, treats NSNumber and TimeInterval as valid sources for the interval, and uses the default reminder time when the value is missing, unrecognized, or not finite. The main flow calls that helper so the reminder time and localized body stay consistent without duplicating parsing logic.

## Verification

On a Mac with Xcode and the project’s `grace` CLI, run `grace ci` (or `grace test` per your usual workflow) to lint and simulator-build. Manually confirm behavior by enabling daily reminders, changing the reminder time, and verifying the next scheduled notification reflects the chosen time after foregrounding or other triggers that reschedule the reminder.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Services/Reminders/DailyReminderNotificationSync.swift
+++ b/GraceNotes/GraceNotes/Services/Reminders/DailyReminderNotificationSync.swift
@@ -13,11 +13,7 @@ enum DailyReminderNotificationSync {
         let status = await reminderScheduler.currentReminderStatus()
         guard status == .enabled else { return }
 
-        let stored = userDefaults.object(forKey: ReminderSettings.timeIntervalKey) as? TimeInterval
-        var interval = stored ?? ReminderSettings.defaultTimeInterval
-        if !interval.isFinite {
-            interval = ReminderSettings.defaultTimeInterval
-        }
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

**Update:** This round does not add new reviewer feedback to respond to, and there are no file changes in this pass to summarize.

If the next step is a code change, I will describe the user-visible behavior, why we made it, and how it was verified (for example with `grace test` or `grace ci` on a Mac) in a follow-up comment tied to that commit.

PR #457, opened 2026-04-14
Honor app UI language for share-card CJK typography

kipyin:
## Headline

Share cards now pick CJK typography when the app is in Chinese, Japanese, or Korean—even if the device language is English.

## User impact

Share images and typography now match the language you chose in the app instead of only following the system language. That makes shared journals look correct for localized users who keep English as their device language.

## What changed

Share typography script selection used to rely on the system locale alone, so CJK styling could be wrong when the app UI was set to Chinese, Japanese, or Korean while the device stayed on English. The resolver now looks at the app bundle’s preferred localizations first, then the current locale, and treats zh, ja, and ko as CJK. A small helper centralizes that language-code check.

## Verification

On a Mac with Xcode and the project’s grace CLI: run `grace ci` (or `grace test` with the usual simulator destination) after building Grace Notes; manually verify a share preview or export with app language set to zh/ja/ko and system language English to confirm CJK typography applies.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Services/ShareTypographyScript.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Services/ShareTypographyScript.swift
@@ -8,12 +8,23 @@ enum ShareTypographyScript: Equatable, Sendable {
 }
 
 extension ShareTypographyScript {
-    static func current() -> ShareTypographyScript {
-        switch Locale.current.language.languageCode {
-        case .some(.chinese), .some(.japanese), .some(.korean):
+    /// Consults the app bundle’s active UI language before the system locale, so CJK typography
+    /// applies when the app is localized to Chinese/Japanese/Korean even if the device language is English.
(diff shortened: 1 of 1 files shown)


kipyin:
**Update:** Share typography now treats the app’s **active** UI language as `bundle.preferredLocalizations.first` only (with optional overrides for tests), so we no longer scan the full preferred list and accidentally pick CJK fonts when English is the real UI language but another language appears later in the user’s preferences. `current(bundle:preferredUILocalizationIdentifier:systemLanguage:)` checks that UI identifier first, then falls back to the system language, and `resolvedBaseLanguageCode` tightens CJK detection when `languageCode` is missing by using the locale wrapper and the leading BCP‑47 tag. New `ShareTypographyScriptTests` lock in precedence (including English UI + Japanese system vs English UI + English system), `zh-Hans` / `zh-Hant` / `ja` / `ko` ahead of English system, and Latin when both UI and system are non-CJK.

PR #460, opened 2026-04-14
Stabilize weekly review recurring themes and tighten theme aggregation

kipyin:
## Headline

Weekly review chips pick clearer winners when themes tie, with more reliable aggregation under the hood.

## User impact

Recurring gratitude, need, and people themes in the weekly review are less likely to flip arbitrarily when counts match. The change also makes theme rollups behave more predictably when the same theme appears again in a week.

## What changed

Recurring theme selection for the weekly review now breaks ties using the order a theme first appeared in the scanned week, aligned with how full theme summaries are ranked. Theme accumulation was rewritten so updates merge into an existing normalized theme in one pass instead of relying on repeated optional lookups and early returns.

## Verification

On macOS, run `swiftlint lint` from the repo root and `grace ci` (or `grace test` with the project’s default simulator destination) to confirm lint and the GraceNotes scheme build/tests pass.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Services/WeeklyReviewAggregates.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Services/WeeklyReviewAggregates.swift
@@ -224,6 +224,9 @@ private extension WeeklyReviewAggregatesBuilder {
                 if $0.dayCount != $1.dayCount {
                     return $0.dayCount > $1.dayCount
                 }
+                if $0.firstSeenOrder != $1.firstSeenOrder {
+                    return $0.firstSeenOrder < $1.firstSeenOrder
+                }
                 return $0.displayLabel.localizedCaseInsensitiveCompare($1.displayLabel) == .orderedAscending
             }
             .prefix(maxThemesPerSection)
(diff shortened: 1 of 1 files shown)


kipyin:
Thanks for the careful read. `topThemes` now breaks mention/day ties the same way chip-only summaries do: `firstSeenOrder` (journal and chip scan order) comes before a case-insensitive label sort, so recurring gratitudes/needs/people chips stay stable and aligned with `sortedThemeSummaries` instead of reshuffling on alphabet alone. The `accumulateTheme` path now does a single read/modify/write merge for existing keys, which matches the old insert-then-update behavior but is easier to follow. I added `WeeklyReviewChipOrderTests` with a fully tied four-chip week (non-alphabetical labels) to assert the top-three recurring chip order and to assert `recurring*` matches the first three `ThemeSummary` rows from `candidateInputs` for gratitudes, needs, and people—so a future sort tweak should fail tests rather than silently reorder chips. Full validation is still `grace test` / `grace ci` on macOS because XCTest needs the simulator toolchain here.

PR #461, opened 2026-04-14
Respect Reduce Motion for themed button press scaling

kipyin:
## Headline

Press feedback stays calm for people who reduce interface motion.

## User impact

Users who enable Reduce Motion in iOS accessibility settings avoid a subtle shrink-on-press effect that can feel unnecessary or uncomfortable. Taps still read as pressed through non-scaling feedback (for example opacity), so affordance remains without extra motion.

## What changed

A Warm Paper button style in the design system applied a small scale-down whenever a control was pressed, even when the user asked the system to limit motion. The press scale is now skipped when Reduce Motion is on, matching the existing policy that already avoided animated easing in that case. This keeps interaction feedback consistent with system accessibility expectations and reduces vestibular or distraction risk from micro-motion.

## Verification

On a Mac with Xcode, run `grace ci` (or `grace test` with the project’s default simulator destination). In the Simulator or on device, turn on Settings → Accessibility → Motion → Reduce Motion, then press themed buttons and confirm they no longer shrink while pressed; with Reduce Motion off, confirm the subtle press scale still appears.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/DesignSystem/Theme.swift
+++ b/GraceNotes/GraceNotes/DesignSystem/Theme.swift
@@ -338,7 +338,7 @@ struct WarmPaperPressStyle: ButtonStyle {
 
     func makeBody(configuration: Configuration) -> some View {
         configuration.label
-            .scaleEffect(configuration.isPressed ? 0.98 : 1.0)
+            .scaleEffect(configuration.isPressed && !reduceMotion ? 0.98 : 1.0)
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

This pass did not change any files or add new commits. There was no reviewer feedback to address in this round, so there is nothing substantive to summarize in the diff. If you expected edits from this pass, say what should be updated and we can handle it in a follow-up.

PR #464, opened 2026-04-14
Fix interface chrome font fallback and disabled bar button styling

kipyin:
## Headline

System chrome typography stays predictable when Outfit is missing, and disabled toolbar buttons match the rest of the UI.

## User impact

If the bundled Outfit font failed to load, navigation bars, tab bars, and bar buttons could end up oversized or inconsistent with Dynamic Type because the fallback was scaled twice. Disabled bar buttons could also look different from enabled ones because only normal and highlighted states used the custom font.

## What changed

The Dynamic Type scaling helper now uses the system text style at a fixed Large content size when Outfit is unavailable, so UIFontMetrics applies a single predictable scaling path instead of compounding the user’s current text size. Global bar button item appearance also sets the same font for the disabled control state so disabled actions stay visually consistent with normal and highlighted titles.

## Verification

On a Mac with Xcode, check out the branch and run `grace ci` (or `grace test` with the project’s default simulator destination). In Simulator, optionally remove or break the Outfit font face to confirm chrome falls back without absurd scaling, and verify disabled bar buttons use the same typography as enabled ones at a few Dynamic Type sizes.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/DesignSystem/AppInterfaceAppearance.swift
+++ b/GraceNotes/GraceNotes/DesignSystem/AppInterfaceAppearance.swift
@@ -16,7 +16,14 @@ enum AppInterfaceAppearance {
         textStyle: UIFont.TextStyle,
         maximumForContentSizeCategory limit: UIContentSizeCategory? = nil
     ) -> UIFont {
-        let baseFont = UIFont(name: name, size: baseSize) ?? UIFont.preferredFont(forTextStyle: textStyle)
+        let baseFont: UIFont
+        if let outfit = UIFont(name: name, size: baseSize) {
+            baseFont = outfit
+        } else {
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

Thanks for the deep review.

We fixed the Outfit fallback so the system font is anchored at the standard text style size (`.large`) before `UIFontMetrics.scaledFont(for:)`, which stops Dynamic Type from scaling the font twice when Outfit does not load. We also set `.disabled` bar button title attributes to mirror the normal and highlighted states so disabled navigation chrome stays visually consistent without overriding colors or other attributes.

We agree there is no practical automated coverage for global `UIAppearance` and `UIFontMetrics` here; a quick Simulator pass with Dynamic Type—and optionally with Outfit unavailable—is the right sanity check. Full validation should be run on a Mac with `grace test` or `grace ci` per project workflow; the automated review environment did not run those.

PR #465, opened 2026-04-14
Harden journal share image export scale and bitmap validation

kipyin:
## Headline

Sharper, more reliable share-card images when sheets change scene state, with stricter bitmap checks.

## User impact

Share exports are less likely to look soft or low-resolution when a sheet leaves the main window scene in a foreground-inactive state. The renderer also refuses images that lack a usable backing bitmap, avoiding odd or empty-looking shares.

## What changed

The share card bitmap path now requires a non-nil CGImage with positive pixel dimensions, instead of only validating dimensions when a CGImage happened to exist. Pixel scale selection for ImageRenderer now considers both foreground-active and foreground-inactive window scenes (still excluding background and unattached scenes), so a presented sheet does not cause the code to miss the on-screen Retina scale and fall back to weaker sources. Comments were updated to describe this behavior.

## Verification

On a Mac with Xcode and simulators, run `grace ci` (or `grace test` with the project’s default destination) and manually share a journal entry with a sheet presented so the main scene is foreground inactive; confirm the exported image looks sharp and sharing still succeeds.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Services/JournalShareRenderer.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Services/JournalShareRenderer.swift
@@ -14,9 +14,8 @@ enum JournalShareRenderer {
         renderer.scale = Self.recommendedPixelScale()
         guard let image = renderer.uiImage else { return nil }
         guard image.size.width > 0, image.size.height > 0 else { return nil }
-        if let cgImage = image.cgImage {
-            guard cgImage.width > 0, cgImage.height > 0 else { return nil }
-        }
+        guard let cgImage = image.cgImage else { return nil }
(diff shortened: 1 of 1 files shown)


kipyin:
Updated the doc comment on `recommendedPixelScale()` so it matches the real behavior: we take the maximum positive scale from the current trait collection, any foreground window scene (active or inactive, so a sheet over the journal does not drop scale), and the main screen, then clamp between 1× and 3×, and we skip background or unattached scenes so multi-window setups do not follow another display. Relaxed the `UIImage` guard so we only reject when a backing `CGImage` exists but has zero dimensions; if there is no `cgImage` but the image has a positive size, we still use it for sharing. Added a short inline note that using `max()` across foreground scenes is intentional so export uses the sharpest attached screen when several windows are open.

PR #468, opened 2026-04-14
Treat empty fullText as missing when decoding journal entries

kipyin:
## Headline

Imported or legacy journal lines no longer lose their text when fullText is empty but labels still hold the real content.

## User impact

Some exports or older data can store the visible line in legacy label fields while fullText is present but empty. Before this change, those lines could decode as blank. After the change, empty fullText is ignored so the stored labels are used and the line shows up as expected.

## What changed

Journal item decoding now treats a present but empty fullText the same as a missing fullText, so the decoder falls back to legacy entryLabel and chipLabel when they carry the real text. The comment was updated to describe that legacy rows may omit fullText, use an empty fullText, or only use the older label fields.

## Verification

On a Mac with Xcode and the project dev tools installed, run `grace ci` (or `grace test` with the usual simulator destination) to confirm the GraceNotes scheme still builds and tests pass.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Data/Models/Entry.swift
+++ b/GraceNotes/GraceNotes/Data/Models/Entry.swift
@@ -15,10 +15,10 @@ struct JournalItem: Codable {
         id = try container.decodeIfPresent(UUID.self, forKey: .id) ?? UUID()
         let entryLabel = try container.decodeIfPresent(String.self, forKey: .entryLabel)
         let chipLabel = try container.decodeIfPresent(String.self, forKey: .chipLabel)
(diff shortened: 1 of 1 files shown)


kipyin:
Added **JournalItemDecodingTests** so legacy JSON where `fullText` is an empty string but `entryLabel` or `chipLabel` holds the real text decodes to that label in `JournalItem.fullText`, matching the fix in `Entry`/`JournalItem` decoding. The suite covers entry-only, chip-only when entry is absent, and entry preferred over chip when both are present. This guards against regressions where an empty exported `fullText` would otherwise produce a blank strip and drop recoverable label text.

PR #469, opened 2026-04-14
Show loading skeleton when Most Recurring refresh has empty cached themes

kipyin:
## Headline

The Most Recurring review card no longer looks blank under the title while fresh insights load.

## User impact

During a refresh, the app could still show a cached snapshot with no recurring themes, which left the card title visible but the body empty and easy to mistake for missing data or a bug. The card now shows the same loading placeholders used elsewhere so the screen clearly signals that results are on the way.

## What changed

The Most Recurring card already stayed on screen while loading so the section title did not disappear with stale empty data. That behavior could still render an empty content area under the title when loading was true but the cached weekly stats had no themes. The view now treats that case like an in-progress load: it swaps the empty panel for the loading skeleton until non-empty themes arrive or loading finishes, matching the loading experience described for similar review insight cards and improving perceived reliability.

## Verification

On a Mac with Xcode and the project’s grace CLI, run `grace ci` (or `grace test` with the default simulator destination from gracenotes-dev.toml). Manually, open Review, trigger a refresh when the prior snapshot had no recurring themes, and confirm the Most Recurring card shows loading placeholders under the title instead of a blank panel.

## Risk

Medium

## Touch class

`ui-ux`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Views/ReviewMostRecurringCard.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Views/ReviewMostRecurringCard.swift
@@ -44,7 +44,12 @@ struct ReviewMostRecurringCard: View {
         if isLoading, insights == nil {
             loadingSkeleton
         } else if let insights {
-            contentForInsights(insights)
+            // Avoid an empty panel under the title when a refresh still has a cached empty snapshot.
+            if isLoading, insights.weekStats.mostRecurringThemes.isEmpty {
+                loadingSkeleton
+            } else {
+                contentForInsights(insights)
+            }
         }
     }
 


kipyin:
**Sentry — review feedback**

**What changed and why**

We adjusted `ReviewMostRecurringCard` so that when insights are still loading (`isLoading`) and we already have a non-nil `insights` value but `mostRecurringThemes` is empty, the card shows the loading skeleton instead of the normal “content for insights” body. That matches how `ReviewScreen` works: `reviewInsights` can stay the previous/cached snapshot while `generateInsights` runs, so without this branch the UI could briefly show a misleading empty state during refresh.

`shouldShowCard` is unchanged: once loading finishes, if there are still no themes, the card stays hidden as before, so we don’t get a stuck “title + empty body” panel. I haven’t added automated tests for the cached-empty-while-refreshing path; validating that stays manual or would need a snapshot/UI test that forces the timing. I’ll run **`swiftlint lint`** and **`grace ci`** on a Mac per the project docs before merge.

PR #472, opened 2026-04-14
Stabilize share card row identity and fix share card layout

kipyin:
## Headline

Journal share cards identify rows and sections reliably so toggling content stays smooth and aligned.

## User impact

Using list position as the only identity made SwiftUI reuse the wrong views when sections or lines changed, which could cause flicker, odd animations, or taps hitting the wrong line. Section titles and read-only body text now lay out consistently with the rest of the card so shared previews look even and predictable.

## What changed

Sections in the share card ForEach are keyed by section kind instead of index. Each line row gets a stable id: visible and redacted lines use their share line identity, and preview placeholder lines use the section kind plus index so stubs stay distinct without colliding with real lines. The non-interactive section title path now allows vertical growth like the toggle variant, and non-tappable body text uses full-width leading alignment to match interactive lines.

## Verification

On macOS with Xcode and the iOS Simulator, run `grace ci` (or `grace test` with the repo’s default destination). In the app, open share preview and toggle sections and individual lines; confirm rows update without visual glitches, and that taps still target the intended line.

## Risk

Medium

## Touch class

`ui-ux`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Views/JournalShareCardView.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Views/JournalShareCardView.swift
@@ -35,6 +35,16 @@ struct JournalShareCardView: View {
     }
 }
 
+private enum ShareLineRowID: Hashable {
+    case line(ShareLineIdentity)
+    case previewStub(section: ShareSectionKind, index: Int)
+}
+
+private struct IdentifiedShareLineRow: Identifiable {
+    let id: ShareLineRowID
+    let item: ShareLineDisplayItem
+}
+
 private extension JournalShareCardView {
     var surface: ShareCardSurface { payload.cardSurface }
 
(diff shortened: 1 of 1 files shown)


kipyin:
Thanks for the thorough pass—both points made sense.

**Section `ForEach` identity:** I added a short invariant comment next to the section `ForEach` noting that `payload.sections` must not repeat the same `ShareSectionKind`, which matches how `ShareRenderPayloadBuilder` builds the list today. I also added a **debug-only** assertion in `ShareRenderPayloadBuilder` so duplicate kinds fail fast in development instead of surfacing as confusing SwiftUI identity issues.

**Stub line layout:** I applied the same `.frame(maxWidth: .infinity, alignment: .leading)` treatment to `stubShareLine` as `visibleShareLine`, so placeholder / excluded-section rows expand and align like body lines in narrow or flexible-width layouts.

I’ll still spot-check share-card export in the simulator for a few representative payloads, since we don’t have automated snapshot coverage for this path.

PR #473, opened 2026-04-14
Harden journal history month grouping for calendar edge cases

kipyin:
## Headline

Journal history stays grouped by month even when the system calendar cannot build a clean month start.

## User impact

Entries that belong in the same calendar month are less likely to appear under multiple month headers or get scattered by day. The change also avoids a fragile assumption about UTC time zone construction, which keeps month-key logic reliable across environments.

## What changed

The Gregorian UTC fallback used when the active calendar cannot form a month start now picks a UTC-like time zone with safe fallbacks instead of force-unwrapping. When building the first day of the month from components fails but year and month are still known, the code derives that month from January 1 by adding months, and only then falls back to start-of-day. That reduces cases where the previous fallback split one month across several buckets. Documentation comments were updated to describe this ordering and the remaining last-resort behavior.

## Verification

On a Mac with the repo’s grace CLI installed, run `grace ci` (or `grace test` with the project’s default simulator destination) to lint and build; optionally open Journal history and confirm entries for unusual dates or calendar settings still appear under a single month section.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/HistoryEntryGrouping.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/HistoryEntryGrouping.swift
@@ -38,11 +38,15 @@ enum HistoryEntryGrouping {
     }
 
     /// When the active calendar cannot form a month start, anchor by Gregorian UTC year/month
-    /// so entries in the same month are not split by day. `startOfDay` is only used if even
-    /// this normalization fails.
+    /// so entries in the same month are not split by day. If direct month normalization fails
+    /// but year and month are known, derive the month from January 1; only then fall back to
+    /// `startOfDay(for:)` (which can still split one month across buckets when all else fails).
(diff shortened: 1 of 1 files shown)


kipyin:
**What changed**

Month keys for journal history now stay stable when the normal calendar path fails: we still try `dateInterval` and `date(from:)` first, then fall back to a Gregorian UTC path that can anchor from January 1 and add months when `date(from:)` cannot build the first of the month but year and month are known—so one calendar month does not get split across many day buckets. The January‑1 + month-add logic lives in `gregorianUTCMonthStartFromYearMonth`, marked `internal` so tests can hit that path without relying on brittle calendar setups. A new unit test calls that helper directly and documents why we do not try to force the full private chain across OS versions. We also note that `clampedMonth` is an ultra-defensive guard so bad component values do not crash. Please run `grace ci` on a Mac to confirm the suite and build match project CI.

PR #474, opened 2026-04-14
Document streak refresher contract for full journal history

kipyin:
## Headline

Streak math stays accurate when callers pass the right data—now the code says so.

## User impact

If streak summaries were ever computed from a partial list of days, the app could show a shorter streak than the user actually has. This change does not alter runtime behavior; it documents the requirement so maintainers avoid that mistake when reusing the in-memory path.

## What changed

The streak summary refresher’s documentation was tightened. The overload that takes journals already loaded from the store now states that callers must pass every journal row that could affect streak continuity—typically the same full set as the repository’s fetch-all method—and explains that filtering to only recent days can shrink the computed streak. The repository-based overload is documented as the path that loads all rows, with a note to prefer the in-memory overload when data is already available. Terminology was aligned (journal list vs. entry list) for clarity.

## Verification

Review the updated doc comments in JournalStreakSummaryRefresher.swift. On a Mac with Xcode, run `grace ci` (or `grace test`) to confirm the project still builds and tests as before; this change is documentation-only.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Services/JournalStreakSummaryRefresher.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Services/JournalStreakSummaryRefresher.swift
@@ -2,7 +2,11 @@ import Foundation
 import SwiftData
 
 enum JournalStreakSummaryRefresher {
-    /// Computes streak summary from an entry list already loaded from the store (avoids a second full fetch).
+    /// Computes streak summary from a journal list already loaded from the store (avoids a second full fetch).
+    ///
+    /// - Parameter entries: Must include every journal row that could affect streak continuity—typically the same
+    ///   set as ``JournalRepository/fetchAllEntries(context:)``. Passing a filtered subset (for example only recent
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

Thanks for the deep review and for calling out the residual risks.

This PR only updates documentation on `JournalStreakSummaryRefresher`: it states that `loadSummary(calculator:entries:now:)` must be given the **full** set of journal entries that can affect streak continuity—the same idea as feeding `StreakCalculator` a complete picture of completed days—so callers don’t pass a subset and get an understated streak. That aligns the refresher’s contract with how `StreakCalculator` walks backward from “today” and treats missing completed days as gaps.

We’re not adding compile-time enforcement here; call sites that use the in-memory overload still need to pass a complete list (as `JournalRepository`’s path does via `fetchAllEntries`). A targeted test tying “subset understates streak” to this API could be a follow-up if we want extra guardrails; a quick DocC build is reasonable if we publish API docs and want to verify symbol links.

PR #476, opened 2026-04-14
Fix review theme evidence decoding for legacy cached payloads

kipyin:
## Headline

Weekly review insights load reliably when cached data still uses the older recurring-theme evidence format.

## User impact

Users who upgraded with older cached review data could hit decoding failures or lose recurring-theme evidence because the stored JSON shape no longer matched the newer type. The app now tolerates both shapes so Review stays stable across updates and older on-device caches.

## What changed

The most recurring theme decoder treated legacy evidence as optional with decodeIfPresent, but when the evidence key exists with the legacy array shape, decoding the new type throws instead of returning nil, which can abort loading the whole cached review payload.

Evidence decoding now lives in a helper that tries the structured form and then the legacy form using try? full decodes, maps legacy rows into surface evidence (deduplicating sources per row), and falls back to an empty list when neither works. A short comment documents why decodeIfPresent is insufficient.

## Verification

From the repo root, run swiftlint lint. On a Mac with Xcode 26+, run grace ci (or grace test with the default simulator destination from gracenotes-dev.toml) to confirm the GraceNotes scheme builds and automated tests pass.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Services/ReviewInsights.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Services/ReviewInsights.swift
@@ -116,22 +116,27 @@ struct ReviewMostRecurringTheme: Equatable, Hashable, Sendable, Codable, Identif
         dayCount = try container.decode(Int.self, forKey: .dayCount)
         currentWeekCount = try container.decode(Int.self, forKey: .currentWeekCount)
         previousWeekCount = try container.decode(Int.self, forKey: .previousWeekCount)
-        if let structuredEvidence = try container.decodeIfPresent(
-            [ReviewThemeSurfaceEvidence].self,
-            forKey: .evidence
-        ) {
-            evidence = structuredEvidence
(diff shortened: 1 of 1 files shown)


kipyin:
We updated `ReviewMostRecurringTheme` so cached JSON still loads when `evidence` is stored as the older `[ReviewThemeEvidence]` array: we no longer rely on `decodeIfPresent` for that field, because when the key exists but the type does not match, Swift throws and the whole decode failed, which could crash on load. Evidence decoding now treats a missing `evidence` key as an empty list, tries the current `[ReviewThemeSurfaceEvidence]` shape, then maps legacy rows into surface evidence, and throws a clear `DecodingError` if the key is present but neither shape matches, so we do not silently drop data. That matches the “fail when `.evidence` exists but is invalid” option from review. Adding a small `JSONDecoder` round-trip test for the legacy payload and watching `ReviewMovementTheme` (still synthesized `Codable`) are reasonable follow-ups; on a Mac, `swiftlint lint` and `grace ci` remain the checks to run for this change.

PR #477, opened 2026-04-14
Block entry-chip switches when a full section cannot save a draft

kipyin:
## Headline

Prevents losing unsaved composer text when a sequential section is at its slot limit.

## User impact

If you had text in the entry field while no row was being edited and the section already had the maximum number of items, tapping another chip could still move focus and replace the field, which dropped text you had not yet saved. The app now keeps you on the current draft until there is room to commit it or you clear it.

## What changed

The entry-tap handler now treats “draft text with nothing selected to update” the same as a failed save when the section is at capacity: it does not advance to the tapped row’s text. The logic is split so a full section sets switching off instead of falling through with switching still allowed, and failed immediate adds also block switching for consistency.

## Verification

On a Mac with Xcode, run `grace ci` (or `grace test` with the project’s default simulator). Manually fill a sequential section to the maximum slots, type new text without selecting a row, then tap another chip; the draft should remain and focus should not load the other row’s content over it.

## Risk

Medium

## Touch class

`ui-ux`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Views/JournalScreenEntryHandling.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Views/JournalScreenEntryHandling.swift
@@ -163,11 +163,16 @@ enum JournalScreenEntryHandling {
             } else {
                 canSwitch = false
             }
-        } else if !trimmed.isEmpty, operations.count < JournalViewModel.slotCount {
-            if let newIndex = operations.addImmediate(input.wrappedValue) {
-                input.wrappedValue = ""
-                _ = newIndex
+        } else if !trimmed.isEmpty {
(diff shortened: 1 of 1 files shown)


kipyin:
I added a regression test for the at-capacity case where no row is being edited, the inline field has non-empty trimmed text, and the user taps another strip: `performEntryTap` must not switch to that row or load its text—`input` and `editingIndex` stay unchanged, which locks in the `canSwitch = false` path and prevents silent draft loss. I reworded the inline comment so it plainly says the draft cannot be saved before switching and that switching would discard it. `performEntryTap` may still return `true` when no row switch happens; callers should keep treating that as “handled,” not “switched to `tapIndex`.”

PR #478, opened 2026-04-14
Harden journal entry editing with idiomatic bounds checks

kipyin:
## Headline

Journal list edits now validate indices consistently and update rows without redundant work.

## User impact

Edits to gratitudes, needs, and people stay predictable when the UI passes an index, which reduces the chance of crashes or odd saves during reordering or rapid edits. Reordering rows uses clearer insert-position validation so moves stay in range.

## What changed

Entry update and remove paths now use Collection.indices.contains for bounds checks instead of hand-rolled lower and upper comparisons, and immediate and async updates read the existing row once, trim afterward, and reuse its stable id when replacing text. The private move helper uses the same style of checks for the source index and for destination offsets from zero through the array length (valid drop positions), and the full-text helpers align with the same pattern. Behavior for valid user actions should match before, with more consistent guards and slightly clearer ordering of validation.

## Verification

On a Mac with Xcode, run `grace ci` (or `grace test` with the project’s default simulator destination) and exercise editing, deleting, and dragging rows in gratitudes, needs, and people to confirm saves and moves behave as before.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/ViewModels/JournalViewModel+EntryEditing.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/ViewModels/JournalViewModel+EntryEditing.swift
@@ -33,33 +33,39 @@ extension JournalViewModel {
 
     /// Returns true if the item was updated (valid index and trimmed text non-empty).
     func updateGratitude(at index: Int, fullText: String) async -> Bool {
+        guard gratitudes.indices.contains(index) else { return false }
         let trimmed = fullText.trimmingCharacters(in: .whitespacesAndNewlines)
-        guard index >= 0, index < gratitudes.count, !trimmed.isEmpty else { return false }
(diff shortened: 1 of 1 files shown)


kipyin:
**Reply to review**

Thanks for the deep review. I tightened the docstrings on `updateGratitude`, `updateNeed`, and `updatePerson` so they match the real contract: `false` for an invalid index or empty trimmed text, and `true` when the stored value already matches the trimmed input or was updated—so nobody reads `true` as “a write definitely happened.” Implementation and guard ordering are unchanged; this is documentation only. I did not add new tests for the refactored bounds checks because behavior matches the previous `Int` index checks on arrays. Validation on my side remains SwiftLint from the repo root; full `grace test` / `grace ci` should run on a Mac with Xcode as usual.

PR #479, opened 2026-04-14
Stabilize journal completion info card bloom animation lifecycle

kipyin:
## Headline

Cancels the badge info “bloom” animation cleanly so dismissals and badge switches don’t leave stray motion or stuck visuals.

## User impact

Tapping completion badges or closing the info card should feel more predictable, with fewer odd flashes or half-finished highlight effects when you move quickly between badges or dismiss the sheet. Reduce Motion mode still shows a steady completed state without leftover animation work running in the background.

## What changed

The presentation model now routes bloom cancellation through a single helper that stops the bloom task, clears the task handle, and resets bloom progress when that makes sense (for example when the card is dismissed or the view goes away). Dismiss, close-then-reopen, and cleanup paths call this before other work so a delayed bloom step cannot fire after the UI has moved on. For Reduce Motion, the code clears the bloom task without forcing progress back to zero before setting the final value, so the accessibility path stays consistent.

## Verification

On a Mac with Xcode, run `grace ci` (or `grace test` with the project’s default simulator destination). Manually exercise the journal completion badges: open the info card, dismiss it quickly, switch between different badges while the card is open, and repeat with Reduce Motion enabled in Settings to confirm the highlight behaves calmly and does not flicker or stall.

## Risk

Medium

## Touch class

`ui-ux`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Views/JournalCompletionInfoPresentation.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Views/JournalCompletionInfoPresentation.swift
@@ -39,6 +39,7 @@ final class JournalCompletionInfoPresentation {
     }
 
     func dismissInfoCard(reduceMotion: Bool) {
+        cancelInfoCardBloomAnimation(resetProgress: true)
         infoCardDismissTask?.cancel()
         infoCardDismissSequence += 1
         infoCardDismissTask = nil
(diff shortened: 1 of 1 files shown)


kipyin:
When the completion info card is closed and immediately reopened (same badge, rapid taps), we now call bloom cancellation with **`resetProgress: !reduceMotion`** in **`scheduleInfoCardCloseThenReopenAfterDelay`**. With Reduce Motion on, the pill still drives outline opacity via bloom progress while scale stays frozen; resetting progress to zero right before the 1 ms reopen was making the stroke briefly fade out and back in. Skipping that reset in Reduce Motion keeps the outline steady and matches the intended “no gratuitous motion” behavior. Dismiss and disappear paths still reset bloom progress when we tear the card down so delayed tasks cannot keep animating after the UI is gone.

PR #481, opened 2026-04-14
Align weekly continuity prompt with the resurfacing headline insight

kipyin:
## Headline

Weekly review follow-up prompts now stay tied to the same insight as the resurfacing line.

## User impact

When the weekly review shows a specific resurfacing line, the continuity prompt is more likely to match that same thread instead of jumping to a different insight’s action. That reduces confusing mismatches between what you read and what you are asked to reflect on next, and keeps the flow feeling intentional.

## What changed

The engine used to pick the resurfacing headline from the first insight with a usable observation or theme, but still chose the continuity prompt from whichever insight had the first non-empty action—often a different card. The change tracks which insight actually drives the headline and prefers that insight’s trimmed action for the continuity prompt, falling back to any other non-empty action and then the default only when needed. Small helpers centralize trimming and selection so the headline and prompt stay aligned by construction.

## Verification

On a Mac with the project dev tooling installed, run `grace ci` (or `grace test` with the repo’s default simulator destination) to lint and build; optionally exercise the weekly review UI with multiple insights to confirm the follow-up matches the highlighted resurfacing line.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Services/WeeklyInsightRuleEngine.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Services/WeeklyInsightRuleEngine.swift
@@ -48,13 +48,14 @@ struct WeeklyInsightRuleEngine {
 
         let narrativeSummary = candidateBuilder.narrativeSummary(from: selectedInsights)
         let starterReflection = String(localized: "review.insights.starterReflection")
-        let trimmedHeadline = Self.firstNonEmptyTrimmedHeadline(in: selectedInsights)
-        let resurfacingMessage = trimmedHeadline ?? starterReflection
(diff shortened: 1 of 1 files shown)


kipyin:
Added focused unit tests for `continuityPrompt(matching:in:defaultPrompt:)` so we lock in the behavior reviewers asked for: the prompt uses the trimmed action from the same insight that drives the resurfacing headline when two insights disagree, falls back to the first non-empty trimmed action in list order when the primary’s action is missing or only whitespace, and uses the default prompt when every action is empty. Refactored the fallback path to route actions through `nonEmptyTrimmed` (via `compactMap`) so trimming and “empty” rules stay consistent with the primary branch and with `headlineAndPrimaryInsight`, which already uses that helper for observations and themes. This reduces drift risk and documents the pairing in tests instead of relying only on weekly-review manual checks.

PR #484, opened 2026-04-15
Fix add-morph composer state when the last sentence slot is filled

kipyin:
## Headline

Keeps the “add sentence” composer flag and header taps aligned when a section is full or the morph row is hidden.

## User impact

After adding the final item in a capped section, the UI no longer behaves as if the add composer were still open when it is not shown. Tapping the section header also avoids acting on a stale “composing” state when there is no add row.

## What changed

The change tightens when the section header treats an open add morph as real: the commit path runs only when the morph add slot is still available, not merely when a binding says the composer is visible.

When the list of entries changes and the section reaches its slot limit (or becomes empty), the add-morph visibility flag is cleared so it cannot stay true after the morph row disappears. That keeps local behavior and higher-level screen state (for example journal keyboard scroll targets) consistent.

An unused input accessibility label helper was removed as dead code.

## Verification

On a Mac with Xcode and the project’s grace CLI: run `grace ci` (or at least `grace test` with the default simulator destination from gracenotes-dev.toml). Manually: open a sequential section with a slot cap, expand the add morph, add entries until the section is full, then tap the section title and confirm nothing odd happens; confirm keyboard/scroll behavior still matches expectations when adding in non-full sections.

## Risk

Medium

## Touch class

`ui-ux`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Views/SequentialSectionPrimaryColumn.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Views/SequentialSectionPrimaryColumn.swift
@@ -79,14 +79,6 @@ extension SequentialSectionPrimaryColumn {
         !isTransitioning && !isLockedByGuidance
     }
 
-    private var inputAccessibilityLabel: String {
-        String(
-            format: String(localized: "accessibility.sectionInputLabel"),
-            locale: Locale.current,
-            title
-        )
-    }
-
     private var ambientGuidanceOpacity: CGFloat {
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

No new review notes to respond to in this round.

## Files touched in this pass

None.

This round is a follow-up note only; there is no additional commit attached. The branch work tightens the sequential section’s add-morph flow so when every slot is filled, the morph composer is not left “on” in a way that conflicts with what is shown, and header taps only commit an add-morph when that affordance is actually visible. An unused localized accessibility label helper was removed as part of that cleanup. Please validate on a Mac with `grace ci` or `grace test` using the project’s default simulator destinations.

PR #485, opened 2026-04-15
Refactor backup folder path resolution with explicit path kinds

kipyin:
## Headline

Backup folder picking now uses one clear check for whether each path is missing, a folder, or blocked by a file.

## User impact

Choosing a backup location should behave the same, with fewer chances for subtle inconsistencies when the Files app reports paths. If something is wrong at a path, you still get a clear failure instead of silent misbehavior.

## What changed

The backup folder resolver used repeated `fileExists` checks in a few places, which was easy to read wrong and could diverge over time. The change introduces a small `PathKind` (missing, directory, or not a directory) and a single helper that classifies a URL once per step. `requireDirectoryIfExists` and `ensureDirectoryExistsOrCreate` now branch on that classification with switches, so the rules for “allowed,” “create,” and “error” line up in one place.

## Verification

On a Mac with Xcode, run `swiftlint lint` from the repo root, then `grace ci` (or `grace test`) to confirm the project still builds and tests pass.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Settings/Services/BackupFolderPickerResolution.swift
+++ b/GraceNotes/GraceNotes/Features/Settings/Services/BackupFolderPickerResolution.swift
@@ -9,6 +9,20 @@ enum BackupFolderPickerResolution {
         case pathComponentIsNotDirectory(URL)
     }
 
+    private enum PathKind {
+        case missing
+        case directory
+        case notDirectory
+    }
+
+    private static func pathKind(at url: URL, fileManager: FileManager) -> PathKind {
+        var isDirectory: ObjCBool = false
+        guard fileManager.fileExists(atPath: url.path, isDirectory: &isDirectory) else {
+            return .missing
+        }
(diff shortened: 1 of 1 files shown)


kipyin:
Thanks for the thorough review. This change pulls backup-folder picker path handling into `BackupFolderPickerResolution` so we standardize the chosen URL, require existing path segments to be directories when present, and create the `Grace Notes Backup` leaf (or use it when it already exists) with a single, explicit flow. If `createDirectory` fails—e.g. because something else created a file at that path between checks—we re-check `pathKind` at the target URL and map the “became a file” case to `ResolutionError.pathComponentIsNotDirectory` instead of surfacing only a raw file-manager error. The edge cases you flagged (non-standard parent with a file at the pick, symlink-heavy layouts, and full security-scoped bookmark / scheduled-backup flows) are unchanged in product behavior and remain best validated on macOS via `GraceNotesTests`; adding a focused unit test for the parent-path file case is a reasonable follow-up if we want that scenario locked in.

PR #486, opened 2026-04-15
Fix weekly review theme labels when themes span multiple sections

kipyin:
## Headline

Weekly review picks the right label style when the same theme appears across Gratitudes, Needs, or People.

## User impact

Most recurring and trending themes read with labels that match where you actually wrote about them, instead of almost always sounding like Gratitudes. That makes family- and needs-oriented phrasing show up when those sections are what drove the theme.

## What changed

Previously, final theme display labels were rebuilt using Gratitudes-style rules for nearly every canonical theme, with a narrow exception for “mom” and “dad.” That meant themes that really surfaced from Needs or People could still get the wrong label tone.

The builder now records which structured section types contributed each canonical theme while scanning entries, then chooses a single display source using a fixed priority (People, then Needs, then Gratitudes, then reading notes and reflections). That chosen source is what `displayLabel` uses, while the mom/dad shortcut still forces People.

Latin word-boundary matching for support text also bails out on an empty needle so we do not build a pointless regular expression in edge cases.

## Verification

On a Mac with Xcode and the project’s `grace` CLI installed, run `grace ci` from the repo root (lint plus simulator build; use `grace test` if you want the test target explicitly). Exercise Weekly Review with entries that repeat the same theme across Gratitudes, Needs, and People and confirm the theme title reads with the expected section-appropriate labeling.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Services/WeeklyReviewAggregatesBuilder+ThemeSections.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Services/WeeklyReviewAggregatesBuilder+ThemeSections.swift
@@ -28,6 +28,8 @@ extension WeeklyReviewAggregatesBuilder {
             forJournalCorpus: journalCorpus
         )
         var map: [String: DistilledThemeAccumulator] = [:]
+        /// Strongest section source seen for each canonical (drives `displayLabel`; `.people` enables person labels).
+        var displaySourceByCanonical: [String: ReviewThemeSourceCategory] = [:]
         var sequence = 0
 
         for entry in entries {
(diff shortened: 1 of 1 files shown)


kipyin:
Thanks for the thorough review. I added a focused XCTest that builds the same canonical theme across multiple structured surfaces (including People, Needs, and Gratitudes) with intentionally different `displayLabel` inputs per `ReviewThemeSourceCategory`, and asserts the weekly review “most recurring” (and, where relevant, trending) labels pick the intended ranked source so priority changes cannot slip past CI unnoticed. I also replaced the local `///` above the “strongest section source” variable with a normal `//` comment so it reads as implementation detail rather than API documentation. For equal-rank ties in `strongerDisplaySourceForThemeLabel`, behavior stays deterministic for a fixed dataset by preserving the first winning category encountered; I’ve called that out inline so future tweaks to tie-breaking are deliberate rather than accidental. I re-ran SwiftLint on the touched files; full behavioral proof should still come from macOS `grace test` / `grace ci` with XCTest.

PR #487, opened 2026-04-15
Align daily reminder copy with streak logic when multiple journals share a day

kipyin:
## Headline

Reminder text now reflects your real progress when more than one journal exists for today.

## User impact

If the same calendar day had multiple journal rows, the app could pick one row for notification wording while streak math treated the day differently. That mismatch could show confusing or discouraging reminder copy. After this change, completion for the notification matches the same “any row counts” idea the streak uses.

## What changed

The builder no longer derives “today’s completion” from a single fetch of one journal for the current moment. It looks at every journal whose day matches today and applies the same ordering as streak logic: bloom counts as complete, meaningful content counts as in progress, otherwise empty. Non-lapse reminder body building is split into a small helper so that path stays clear. A short doc comment explains why aggregation matters when duplicate day rows exist.

## Verification

On a Mac with Xcode and the project’s grace CLI, run `grace ci` (or `grace test` with the default simulator destination from gracenotes-dev.toml). Manually, trigger a daily reminder on a day where you have multiple journal entries and confirm the notification wording matches whether any entry is complete or in progress for that day.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Services/Reminders/ReminderNotificationBodyBuilder.swift
+++ b/GraceNotes/GraceNotes/Services/Reminders/ReminderNotificationBodyBuilder.swift
@@ -25,7 +25,6 @@ enum ReminderNotificationBodyBuilder {
         )
 
         if isLapse {
-            // `completion` / `streakBucket` are unused when `isLapse` is true (`localizationKey` returns early).
             let key = ReminderNotificationBodySelector.localizationKey(
                 isLapse: true,
                 completion: .empty,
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

### Comment body (copy as-is)

This update refines how reminder notification copy picks **today’s completion state** when the user is not in a lapse. Instead of relying on a single journal row from the repository, we now **aggregate every journal row for that calendar day** and derive completion the same way streak logic does (bloom wins, then meaningful content, otherwise empty). That keeps notification wording aligned with streak behavior when more than one row exists for the same day. The non-lapse path is split into a small helper so the lapse branch stays a straight shot to localized copy. **Validation:** not run in this environment (iOS build requires macOS with Xcode); please run `grace test` or `grace ci` locally on a Mac before merge.

PR #494, opened 2026-04-16
Remove duplicate journal rows when importing backups

kipyin:
## Headline

Journal import now deletes extra same-day journal rows so each day stays a single source of truth.

## User impact

If the database ever had more than one journal row for the same calendar day (for example from earlier bugs or edge cases), importing could update one row while leaving stale duplicates behind. That could make the timeline or day views confusing or show outdated content. After this change, import removes those extras so each day matches the imported backup cleanly.

## What changed

The import path already deduplicates entries in the backup file by calendar day. On disk, `fetchEntry` returns one canonical journal per day, but other rows for that day could still exist. When updating an existing day, the service now fetches all journals whose `entryDate` falls in that calendar day and deletes every row except the one being updated.

The per-day deduplication helper was moved into an extension for organization, and a small doc comment was wrapped for line length. Behavior for merge conflict detection and file-side deduplication is unchanged except for this cleanup of duplicate persisted rows during apply.

## Verification

On a Mac with Xcode and the project’s `grace` CLI installed, run `grace ci` (or `grace test` with the default simulator destination from `gracenotes-dev.toml`) so lint and simulator build/tests cover this path. Manually, import a backup on a store that has duplicate rows for one day and confirm only one journal remains for that day after import.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Settings/Services/JournalDataImportService.swift
+++ b/GraceNotes/GraceNotes/Features/Settings/Services/JournalDataImportService.swift
@@ -58,11 +58,11 @@ struct JournalDataImportService {
         if let attrs = try? FileManager.default.attributesOfItem(atPath: url.path),
            let num = attrs[.size] as? NSNumber {
             // Use 64-bit magnitude: `intValue` truncates past 32-bit signed range and can mis-report huge files.
-            let v = num.uint64Value
-            if v > UInt64(Int.max) {
+            let uint64Bytes = num.uint64Value
+            if uint64Bytes > UInt64(Int.max) {
                 return Int.max
             }
-            return Int(v)
+            return Int(uint64Bytes)
         }
         return nil
     }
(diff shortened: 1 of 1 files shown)


kipyin:
We tightened import so backup restore matches `JournalRepository.fetchEntry`’s one-row-per-day rule: before applying file fields for a day we coalesce extra `Journal` rows in that calendar interval onto the canonical row from `fetchEntry`, then delete the rest. If two same-day rows carry different meaningful content, we merge non-canonical text into the kept row (slots and notes with explicit rules, timestamps folded conservatively) so we do not drop edits that only lived on a duplicate. Merge mode with **prefer local** still skips importing file data on conflict days, but we now run the same collapse pass for those skipped days so duplicate rows there are removed without overwriting local content. The survivor remains whatever `fetchEntry` ranks as canonical, not necessarily the row the user last touched; fully corrupt or ambiguous duplicate data could still warrant a dedicated conflict/repair path later. SwiftData integration tests should still add a case that seeds two same-day rows and asserts a single row and expected content after import—worth tracking as follow-up coverage.

PR #495, opened 2026-04-16
Skip empty secondary guidance and editing side effects when locked

kipyin:
## Headline

Journal text sections no longer show blank secondary guidance or run keyboard caret work while onboarding locks editing.

## User impact

Readers see cleaner guidance blocks without an extra empty line when secondary copy is blank. While a section is locked during onboarding, typing-related side effects stay quiet, which reduces unnecessary scrolling and caret adjustments that could feel jumpy or unreliable.

## What changed

Secondary guidance now follows the same rule as the primary message: it only appears when the string is non-empty, so the layout does not reserve space for an invisible line.

Text change handling now treats a locked onboarding section as read-only for editing side effects: multiline scroll callbacks and the caret visibility nudge run only when the section is not locked, while newline counting still updates so state stays consistent.

Newline counting is centralized in a small helper so the view uses one clear implementation instead of repeating filtering logic.

## Verification

On a Mac with Xcode and the iOS Simulator, run `grace ci` (or `grace test` with the project’s usual destination). In the app, open a journal section that can show secondary guidance and confirm an empty secondary string does not add a blank line. With onboarding in a locked state for a text section, edit if applicable and confirm no unexpected scroll or caret nudges fire from that path.

## Risk

Medium

## Touch class

`ui-ux`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Views/EditableTextSection.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Views/EditableTextSection.swift
@@ -54,7 +54,8 @@ struct EditableTextSection: View {
                         .font(AppTheme.warmPaperBody)
                         .foregroundStyle(palette.textPrimary)
                         .fixedSize(horizontal: false, vertical: true)
-                    if let guidanceMessageSecondary, !guidanceMessageSecondary.isEmpty {
+                    if let guidanceMessageSecondary,
+                       !guidanceMessageSecondary.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty {
                         Text(guidanceMessageSecondary)
                             .font(AppTheme.warmPaperBody)
                             .foregroundStyle(palette.textPrimary)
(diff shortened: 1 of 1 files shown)


kipyin:
Thanks for the deep review—agreed there were no blockers.

We updated `EditableTextSection` so secondary guidance only renders when `guidanceMessageSecondary` has visible content: the condition now uses trimming for whitespace and newlines, matching how we treat empty-looking keyboard-hint copy elsewhere, so callers cannot leave an empty-looking paragraph while the guidance block stays open.

While a section is onboarding-locked, `onChange(of: text)` still updates `storedNewlineCount` whenever the binding changes, but it no longer runs multiline scroll or global first-responder caret nudges until editing side effects are allowed again—so we avoid spurious scroll/nudge work during lock and avoid a backlog of “extra line” callbacks right after unlock.

We have not added automated tests for this path; a quick simulator check of locked onboarding, focus, and keyboard scroll is still the right manual follow-up on macOS.

PR #496, opened 2026-04-16
Clear tutorial storage when resetting journal onboarding

kipyin:
## Headline

Onboarding reset now truly starts the guided journal path from a clean slate.

## User impact

If someone reset journal onboarding (for example from Settings or support flows), leftover tutorial milestone flags could still make the app treat the install as already onboarded. That could skip or confuse the guided journal experience right after a reset. Clearing those keys prevents the completion flag from being immediately inferred again from old tutorial state.

## What changed

The journal onboarding reset path now removes all stored journal tutorial keys up front, using the existing helper on `JournalTutorialStorageKeys`, before clearing the onboarding-specific `UserDefaults` entries and launch tracking. A short doc comment was added on `resetAll` to explain that this avoids re-deriving “guided journal complete” from leftover tutorial presence when the explicit completion key is gone.

## Verification

On a Mac with Xcode and the project’s `grace` CLI installed, run `grace ci` (or `grace test` with the usual simulator destination from `gracenotes-dev.toml`) to confirm the app still builds and tests pass.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Tutorial/JournalOnboardingProgress.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Tutorial/JournalOnboardingProgress.swift
@@ -128,7 +128,11 @@ final class JournalOnboardingProgress {
         defaults.removeObject(forKey: JournalOnboardingStorageKeys.legacy051GuidedBranchResolution)
     }
 
+    /// Clears journal onboarding keys and tutorial milestone keys so ``resolvedHasCompletedGuidedJournal`` does not
+    /// immediately re-derive completion from leftover tutorial presence (see ``shouldTreatInstallAsPreviouslyOnboarded``).
     static func resetAll(in defaults: UserDefaults = .standard) {
+        JournalTutorialStorageKeys.removeAllStoredKeys(in: defaults)
+
         let keys = [
             JournalOnboardingStorageKeys.completedGuidedJournal,
             LegacyJournalOnboardingStorageKeys.pending051UpgradeOrientation,


kipyin:
**Sentry — review feedback**

Extended `test_resetAll_clearsGuidedJournalAndSuggestionFlags` so it seeds `JournalTutorialStorageKeys.celebratedFirstSprout`, calls `JournalOnboardingProgress.resetAll`, and asserts that key is gone. That locks in the production behavior where onboarding reset clears tutorial storage first so `shouldTreatInstallAsPreviouslyOnboarded` cannot come back from leftover tutorial flags alone, and the suite will fail if `removeAllStoredKeys` is accidentally dropped from `resetAll`. Left the UITest double-reset and broader first-run keys as-is since they are intentional scope and harmless redundancy per the review.

The user asked me to reply with ONLY the comment body - I'll give that as the final output without the implementation details in the message... Actually they wanted the PR comment from the author explaining what changed. Let me refine:

---

Updated `JournalOnboardingProgressTests` so `test_resetAll_clearsGuidedJournalAndSuggestionFlags` sets a tutorial flag in `UserDefaults` before `resetAll` and asserts it is removed afterward. That regression-test directly covers `JournalTutorialStorageKeys.removeAllStoredKeys` inside `JournalOnboardingProgress.resetAll`, which is the fix ensuring leftover tutorial state cannot make `shouldTreatInstallAsPreviouslyOnboarded` treat the install as finished after a reset. We did not change the redundant UITest call order or widen reset beyond journal onboarding plus tutorial storage, since those were called out as non-blocking and existing product scope.

That's a bit long. Shorten to 2-6 sentences.

---

**Final comment body (2-6 sentences, markdown, no HTML):**

Updated `test_resetAll_clearsGuidedJournalAndSuggestionFlags` to set `JournalTutorialStorageKeys.celebratedFirstSprout` before `JournalOnboardingProgress.resetAll(in:)` and assert it is `nil` afterward. This locks in the behavior that onboarding reset clears tutorial `UserDefaults` via `JournalTutorialStorageKeys.removeAllStoredKeys`, so a future edit cannot drop that wipe without breaking tests and `shouldTreatInstallAsPreviouslyOnboarded` cannot re-derive completion from stale tutorial flags alone. No other production code changes in this follow-up; the UITest double-reset and the fact that `resetAll` still does not clear first-run or legacy heuristics outside journal onboarding + tutorial storage stay as reviewed.

Good. User said no fenced code blocks in the reply - the comment uses backticks for symbol names which is normal markdown.

Reply with ONLY the comment body - so my entire response is just that text.
`</think>`
Updated `test_resetAll_clearsGuidedJournalAndSuggestionFlags` to set `JournalTutorialStorageKeys.celebratedFirstSprout` before `JournalOnboardingProgress.resetAll(in:)` and assert that value is cleared afterward. That directly regression-tests the `JournalTutorialStorageKeys.removeAllStoredKeys` call on the onboarding reset path, so we cannot silently remove the tutorial wipe and still have tests pass while `shouldTreatInstallAsPreviouslyOnboarded` could again infer completion from leftover tutorial flags. No other behavioral changes in this pass; the redundant UITest tutorial call before onboarding reset and the existing scope of what “reset” clears (excluding broader first-run keys) remain as discussed in review.

PR #497, opened 2026-04-16
Sort Past statistics history after filtering, not the full journal list

kipyin:
## Headline

Past statistics no longer sorts every journal before applying the selected history window, which keeps large libraries responsive.

## User impact

Users with many entries outside the chosen Past range should see the same charts and lists as before, with less wasted work when opening Past statistics. Reliability improves indirectly by reducing unnecessary sorting on the main thread during those updates.

## What changed

The history window helper used to sort the entire journal list, then drop entries outside the range. That meant the cost of sorting grew with total library size even when only a small slice was needed. It now filters to the validated date range first and sorts only that subset, using the same ordering rules as before.

## Verification

On a Mac with Xcode and the iOS Simulator, run `grace ci` (or `grace test` with the project’s usual destination) to confirm the app still builds and tests pass.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Services/ReviewHistoryWindowing.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Services/ReviewHistoryWindowing.swift
@@ -17,16 +17,16 @@ enum ReviewHistoryWindowing {
         calendar: Calendar,
         pastStatisticsInterval: PastStatisticsIntervalSelection
     ) -> [Journal] {
-        let sortedAll = sortedEntries(allEntries)
         let historyRange = pastStatisticsInterval.validated.resolvedHistoryRange(
             referenceDate: referenceDate,
             calendar: calendar,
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

Thanks for the deep review.

We changed `entriesInValidatedHistoryWindow` so it applies the validated history window first and sorts only the entries in that window, instead of sorting the whole journal and then filtering. That keeps the same membership and the same final order (ascending by date, with `id` as a tie-break) because `resolvedHistoryRange` does not depend on how entries were ordered in memory. The win is mostly practical: when the window is smaller than the full journal, we avoid sorting entries that never appear in the result. We did not add dedicated equivalence tests for filter-then-sort versus sort-then-filter; existing weekly review and aggregate tests still exercise the window and completion paths. Full verification belongs on macOS via `grace test` / `grace ci`, which remains the merge gate.

PR #498, opened 2026-04-16
Fix review inset panel title meta wrapping and empty VoiceOver hints

kipyin:
## Headline

Past review inset panels show full trailing meta text and avoid pointless VoiceOver hints.

## User impact

Long secondary text next to panel titles is less likely to get clipped when the layout stacks vertically. VoiceOver users no longer hear an empty hint when a tappable title is passed an empty hint string.

## What changed

The stacked fallback layout for the title row now allows the trailing meta line to grow vertically so multi-line secondary copy reads fully instead of being squeezed. The optional title accessibility hint helper only applies VoiceOver hint text when the string is non-empty, so we do not attach meaningless hints.

## Verification

On a Mac with Xcode and the project grace CLI installed, run `grace ci` from the repo root (or `grace test` with the project’s default simulator destination) to lint and build; in the simulator, open Past review flows that use ReviewInsightInsetPanel and confirm long trailing meta wraps; with VoiceOver on, verify tappable titles do not announce a blank hint when the hint is omitted or empty.

## Risk

Medium

## Touch class

`ui-ux`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Views/ReviewInsightInsetPanel.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Views/ReviewInsightInsetPanel.swift
@@ -129,8 +129,13 @@ struct ReviewInsightInsetPanel<Content: View>: View {
 private extension View {
     @ViewBuilder
     func optionalAccessibilityHint(_ hint: String?) -> some View {
-        if let hint, !hint.isEmpty {
-            accessibilityHint(hint)
+        if let hint {
+            let trimmed = hint.trimmingCharacters(in: .whitespacesAndNewlines)
+            if !trimmed.isEmpty {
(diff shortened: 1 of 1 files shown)


kipyin:
Thanks for the thorough pass. This change tightens the review inset panel title row so long titles and trailing meta can share a narrow width without awkward crowding: the horizontal layout uses wrapping, scaling, and `ViewThatFits` fallbacks, and the stacked trailing line uses vertical sizing so multi-line meta reads cleanly. For the tappable title, we only attach a VoiceOver hint when there is real copy after trimming whitespace, so empty strings never become hints. I agree a quick check on a narrow iPhone simulator at default and large Dynamic Type is the right follow-up; full `grace` / Xcode validation is on my side per the repo’s macOS workflow.

PR #499, opened 2026-04-16
Pin guided journal completion false after onboarding reset

kipyin:
## Headline

Resetting journal onboarding now leaves users in a true fresh state instead of silently re-marking the guided journal as complete.

## User impact

When support or developers reset onboarding, people should see the full guided journal flow again. Before this fix, clearing stored keys could let the app infer “already completed” from reminders, first-run, or other signals, so the journal could still behave like a finished onboarding even right after a reset.

## What changed

The journal onboarding reset path clears the usual UserDefaults keys and tutorial milestones, then explicitly stores guided journal completion as false. That stops the resolver from treating a missing key as a signal to run the migration heuristics again and immediately write completion back to true. Documentation on the reset helper was updated to explain why removing the key alone was not enough.

## Verification

On a Mac with Xcode, run `grace test` or `grace ci` from the repo root (or the project’s usual CI profile) to lint, build, and exercise tests against the configured simulator destinations.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Tutorial/JournalOnboardingProgress.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Tutorial/JournalOnboardingProgress.swift
@@ -128,8 +128,10 @@ final class JournalOnboardingProgress {
         defaults.removeObject(forKey: JournalOnboardingStorageKeys.legacy051GuidedBranchResolution)
     }
 
-    /// Clears journal onboarding keys and tutorial milestone keys so ``resolvedHasCompletedGuidedJournal`` does not
-    /// immediately re-derive completion from leftover tutorial presence (see ``shouldTreatInstallAsPreviouslyOnboarded``).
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

We updated `resetAll` so that after clearing onboarding-related keys we write `false` for `completedGuidedJournal` instead of only removing it. Removing the key made it look “unset,” which let the guided-journal completion path re-run the install/onboarding heuristics and immediately persist `true` again—effectively undoing the reset. Setting the key to `false` keeps it present so reads go through the same branch as an explicit false and the migration heuristic does not fire right after a deliberate reset. This matches how `optionalBool` treats a present key versus a missing one. I validated the behavior with the updated `test_resetAll_*` contract; full execution still needs an Xcode run (`grace test` / `grace ci` on a Mac). If we add a production “reset journal onboarding” control later, it should call this same API so users get the same pinned-`false` behavior.

PR #501, opened 2026-04-16
Defer Past Review until tab is selected using task(id:)

kipyin:
## Headline

Past now promotes to the full Review screen the first time you open it, not only on cold start.

## User impact

The Past tab is designed to avoid building the heavier Review experience until you actually visit it. When Today is the default tab, a one-shot async task can run while Past is still unselected and never run again when you switch tabs, so the placeholder could stick or behavior could depend on two overlapping hooks. This makes that first visit to Past reliable for performance and correctness.

## What changed

In `DeferredReviewRoot`, the previous approach paired `onChange` of tab selection with a plain `task` that only runs once on first appearance. That mismatch meant selection changes after launch were not what drove the async work. The change collapses this to a single `task` keyed off whether the Past tab is selected, so the deferral logic runs again when `isSelected` flips from false to true after the default Today tab loads. The redundant `onChange` path was removed so one clear mechanism owns flipping the placeholder to `ReviewScreen` and firing the performance trace when Past is actually opened.

## Verification

On macOS with the repo’s toolchain, run `grace ci` (or `grace test` with the project’s usual simulator destination). Manually: launch with Today selected, then switch to Past and confirm the full Past/Review experience appears on first visit without needing a second tap or relaunch.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Application/GraceNotesApp.swift
+++ b/GraceNotes/GraceNotes/Application/GraceNotesApp.swift
@@ -227,12 +227,9 @@ private struct DeferredReviewRoot: View {
                     .navigationTitle(String(localized: "shell.tab.past"))
             }
         }
-        .onChange(of: isSelected) { _, selected in
-            guard selected, !hasOpenedReviewTab else { return }
-            hasOpenedReviewTab = true
-            PerformanceTrace.instant("ReviewScreen.deferredUntilSelected")
-        }
-        .task {
+        // `.task` alone does not re-run when `isSelected` flips after first load; drive off `id` so Past
+        // defers `ReviewScreen` until the tab is actually selected (default tab is Today).
+        .task(id: isSelected) {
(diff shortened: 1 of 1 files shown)


kipyin:
**What changed**

We fixed the Past tab so the full **Past** experience (the review screen) actually appears the first time you switch to Past. A one-shot `.task` was not re-running when the tab’s selection changed, so the view could stay on the lightweight placeholder instead of promoting to `ReviewScreen`.

**How we addressed the review note**

The “open” gate stays **fully synchronous**: we set `hasOpenedReviewTab` in `openPastTabIfNeeded()` with **no** `await` before that assignment, so nothing async can cancel mid-flight and leave Past stuck on `Color.clear` until you toggle the tab again. Selection is handled with `onAppear` and `onChange(of: isSelected)` calling the same helper.

**Verification**

On macOS we’ll run `swiftlint lint` and `grace ci` per the repo’s workflow before merge.

PR #503, opened 2026-04-16
Harden journal completion morph bloom against invalid progress values

kipyin:
## Headline

Keeps the completion pill’s morph animation stable when progress math goes bad.

## User impact

If bloom progress ever became NaN or infinite (for example from a bad animation handoff or edge-case math), the outline and scale around the completion pill could behave unpredictably or disappear. This change treats those values as zero so the pill stays visible and predictable instead of glitching.

## What changed

The completion pill already clamped morph bloom progress to the 0–1 range for normal values. That clamp did not protect against non-finite floats, which can still flow through min/max and break opacity and scale. The morph bloom progress path now checks finiteness first and falls back to a safe value, then applies the same 0–1 clamp. The comment was updated to describe that behavior in plain terms.

## Verification

On a Mac with Xcode and the project’s grace CLI installed, run `grace ci` (or `grace test` with the usual simulator destination from gracenotes-dev.toml) to confirm the app builds and tests pass. Manually spot-check the journal header completion pill during any morph/celebration transition if you want extra confidence in the visual path.

## Risk

Medium

## Touch class

`ui-ux`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Views/JournalCompletionPill.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Views/JournalCompletionPill.swift
@@ -15,9 +15,11 @@ struct JournalCompletionPill: View {
     var morphAccentColor: Color = .clear
     var morphBloomProgress: CGFloat = 0
 
-    /// Defensive clamp so morph bloom visuals stay stable if progress is driven past 0…1.
+    /// Defensive clamp so morph bloom visuals stay stable if progress is outside 0…1 or non-finite.
     private var clampedMorphBloomProgress: CGFloat {
-        min(max(morphBloomProgress, 0), 1)
+        let raw = morphBloomProgress
+        guard raw.isFinite else { return 0 }
+        return min(max(raw, 0), 1)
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

**Public PR comment (paste as-is)**

We hardened `clampedMorphBloomProgress` so non-finite progress (`nan`, `±infinity`) becomes `0` before we clamp to `0…1`, which keeps `Color.opacity` and `scaleEffect` in a safe range—plain `min`/`max` does not fix `nan` in Swift. That is defensive UI hardening at the pill; it does not change persistence or domain rules. If we ever see non-finite values in logs for `infoCardBloomProgress`, we should fix or normalize them in `JournalCompletionInfoPresentation` so the source stays bounded; until then, the pill stays safe even if a future caller wires a bad value. We did not add an automated test for this edge case because the behavior is trivial, though a regression test would be reasonable if the progress source diversifies. Validation for this path remains on macOS via `grace test` / `grace ci` with the iOS Simulator, not Linux-only toolchains.

PR #505, opened 2026-04-16
Harden journal completion math with clamped section counts

kipyin:
## Headline

Growth-stage labels for your journal now derive from one consistent, defensive count model.

## User impact

Chip-section completion (Soil through Bloom) drives progress cues and history; inconsistent or edge-case counts could skew that story. Clamping counts to zero and expressing bloom, leaf, and twig rules through min/max across sections keeps the stage fair and predictable.

## What changed

Completion and bloom checks used repeated comparisons on raw integers, and twig used easy-to-misread OR/AND logic. The change introduces a small private helper that treats each section count as non-negative, then derives the weakest and strongest section sizes from those values. Bloom and leaf tiers now key off that shared minimum (with a named threshold for the pre-bloom “balanced” bar), and the twig case uses max versus min so “mixed progress” is expressed in plain language. Bloom detection and “minimum across sections” reuse the same path so the app does not drift between call sites.

## Verification

On a Mac with Xcode and the project’s default simulator setup, run `grace ci` (or `grace test` with the usual GraceNotes destination) to lint, build, and exercise automated tests.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Data/Models/Journal.swift
+++ b/GraceNotes/GraceNotes/Data/Models/Journal.swift
@@ -108,9 +108,12 @@ final class JournalEntry {
         needsCount: Int,
         peopleCount: Int
     ) -> Bool {
-        gratitudesCount >= slotCount &&
-            needsCount >= slotCount &&
-            peopleCount >= slotCount
+        let counts = NonNegativeSectionCounts(
+            gratitudesCount: gratitudesCount,
+            needsCount: needsCount,
+            peopleCount: peopleCount
+        )
+        return counts.minCount >= slotCount
     }
 
     /// Minimum count across the three chip sections (weakest section).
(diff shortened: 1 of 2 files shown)


kipyin:
**Response to review**

We routed `hasAtLeastOneEntryInEachSection` through the same non-negative section counts as the static completion helpers so every count-based rule uses one clamping path and future call sites cannot drift. We added focused unit tests for negative and all-negative static inputs, plus a check that `minimumEntryCountAcrossSections` stays aligned with `completionLevel` after clamping, so the `max(0, …)` behavior cannot regress quietly. The twig rule stays `maxCount >= 3 && minCount < 3` with non-negative inputs; `minCount >= slotCount` is unchanged for valid data and still yields bloom when every section meets the cap. We did not change `JournalViewModel.hasReachedBloom` in this pass; if we tighten that later, we should delegate to `Journal.entriesIndicateBloom` to avoid a second definition of bloom. Validation: `grace test` (or `grace ci`) on macOS with the project’s default simulator destinations.

PR #507, opened 2026-04-16
Stabilize App Tour paging and path title metrics

kipyin:
## Headline

Keeps the guided tour on valid pages and measures path titles more reliably.

## User impact

Users who skip the congratulations screen or swipe through the tour are less likely to land on a missing page or see a blank pager state. The growth-path strip should align dots with titles more consistently when layout updates.

## What changed

The tour now clamps the selected page whenever it changes, not only on first appearance, so the pager stays within the first and last visible tabs when the congratulations page is omitted or SwiftUI tries to select an invalid index. A small helper centralizes that range logic for clarity and reuse. The preference key that records each step’s title line height now merges child measurements with the maximum instead of the last value, which better matches multi-line or updating title layout.

## Verification

On a Mac with Xcode, run `grace ci` (or `grace test` with the project’s default simulator destination). Manually open the App Tour with congratulations both shown and skipped; swipe between pages and confirm paging, indicators, and path-strip alignment look correct.

## Risk

Medium

## Touch class

`ui-ux`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Tutorial/AppTourView.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Tutorial/AppTourView.swift
@@ -37,6 +37,11 @@ struct AppTourView: View {
         _iCloudAccountState = StateObject(wrappedValue: ICloudAccountStatusModel())
     }
 
+    /// Keeps `TabView` selection aligned with visible pages (e.g. congratulations page omitted).
+    private func clampedPageIndex(_ index: Int) -> Int {
+        min(max(index, firstPageIndex), lastPageIndex)
+    }
+
     var body: some View {
         VStack(spacing: 0) {
             TabView(selection: $pageIndex) {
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

Thanks for the thorough review. This change fixes two related issues in the app tour: we now clamp `pageIndex` on every update, not only on first appear, so if `TabView` reports a selection that does not match the current page set (for example when the congratulations page is skipped and tag `0` is omitted), the state machine recovers instead of sticking on a missing page. We also changed `AppTourPathTitleMetricsKey.reduce` to take the maximum measured title-line height so a smaller subtree cannot overwrite the preference and throw off dot alignment with the title. I have not run `grace test` / `grace ci` in this session; please validate paging with `skipsCongratulationsPage` on and off (swipes, Next, larger Dynamic Type) on the iOS Simulator when you have a Mac handy.

PR #508, opened 2026-04-16
Derive completion badge info card tint from badge cases

kipyin:
## Headline

Info card accent colors now follow the completion badge cases directly instead of an extra mapping hop.

## User impact

Completion badge help text should look the same as before, with tint colors still matching each growth stage. The change makes styling easier to reason about and safer to maintain if completion levels or mappings evolve.

## What changed

The info card tint helper used to branch on the mapped `JournalCompletionLevel` (soil through bloom). It now switches on `CompletionBadgeInfo` itself (empty through full), which matches how titles and descriptions are already expressed. The color choices stay aligned with the previous mapping: earliest stage muted, light check-in accent, middle stages standard text, and full completion using the full-stage text color.

## Verification

On a Mac with Xcode, run `grace ci` (or `grace test` per project docs) to confirm the GraceNotes scheme builds and tests pass; spot-check the journal completion badge info card in the simulator to confirm tint still matches each stage.

## Risk

Medium

## Touch class

`ui-ux`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Views/CompletionBadgeInfo.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Views/CompletionBadgeInfo.swift
@@ -68,14 +68,14 @@ enum CompletionBadgeInfo: Equatable, Sendable {
     }
 
     func infoCardTintColor(using palette: TodayJournalPalette) -> Color {
-        switch completionLevel {
-        case .soil:
+        switch self {
+        case .empty:
             return palette.textMuted
-        case .sprout:
+        case .started:
             return palette.quickCheckInText
-        case .twig, .leaf:
+        case .growing, .balanced:
             return palette.standardText
-        case .bloom:
+        case .full:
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

**What changed**

We refactored `infoCardTintColor(using:)` so it branches on `CompletionBadgeInfo` instead of `JournalCompletionLevel`. The tint mapping is unchanged: empty still maps to soil, started to sprout, growing/balanced to twig/leaf, and full to bloom—matching the previous behavior tied to completion level.

**Why**

The info card tint should follow the same badge semantics the UI already uses for completion, so the color logic stays aligned with `CompletionBadgeInfo` rather than duplicating the level-to-badge relationship in two places.

**Verification**

SwiftLint is clean on the touched file. I have not run `grace ci` or `grace test` in this session; a quick pass in the Simulator on the date-section info card at each growth stage is the best check for any visual regression.

PR #509, opened 2026-04-16
Generate weekly review insights with a single pass over journal entries

kipyin:
## Headline

Weekly review insight generation does less redundant work when scanning your journal history.

## User impact

Faster, more efficient insight generation when you have many entries, with the same weekly review behavior. You should see snappier review screens and less unnecessary CPU work on device.

## What changed

Building the current- and previous-week entry lists used two full scans of every journal entry—one filter for this week and another for last week. That duplicated work for large libraries. The generator now walks the entry list once, assigns each entry to this week, last week, or neither, and passes those lists into the same analysis as before. Logic for which entries belong to which period is unchanged; only how we collect them is streamlined.

## Verification

On a Mac with the repo and Xcode, run `grace ci` (or `grace test` with the project’s default simulator destination) to confirm the app builds and tests pass after the change.

## Risk

Medium

## Touch class

`business-logic`
---


kipyin:
**Sentry — review feedback**

**Response to review**

This change replaces two separate `filter` passes over the same journal list with one forward pass that buckets each entry into the current period, the previous period, or neither. That keeps the same classification rules as before while cutting redundant work. The periods are defined as half-open ranges that meet at `current.lowerBound`, so an entry still cannot land in both buckets; if those ranges are ever refactored so they could overlap, we should add a test (or adjust bucketing) so we do not double-count. I did not run `grace test` / `grace ci` in this environment; validation on a Mac with Xcode remains the right check for the full `GraceNotes` build and tests.

PR #510, opened 2026-04-16
Make journal export timestamps thread-safe without shared DateFormatter

kipyin:
## Headline

Export filenames stay stable in UTC while avoiding unsafe shared formatters under concurrency.

## User impact

Journal data exports still use the same UTC timestamp pattern in filenames, but the app no longer risks data races from a single shared DateFormatter when multiple exports run at once. That reduces rare crashes or corrupted filenames and makes background or overlapping exports more reliable.

## What changed

The export filename timestamp used a static DateFormatter, which Apple documents as not thread-safe. That becomes a problem if exports are triggered from concurrent work (for example detached tasks) because every call would share the same formatter instance.

The change removes that shared formatter. Timestamps are built with a Gregorian calendar in UTC, POSIX locale, and fixed-width numeric formatting so the string matches the previous yyyyMMdd-HHmmss style without depending on locale. A brief comment documents why this approach is safer under concurrency.

## Verification

On a Mac with Xcode and the project’s grace CLI installed, run `grace ci` (or `grace test` with the usual simulator destination) from the repo root to lint, build, and exercise tests.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Settings/Services/JournalDataExportService.swift
+++ b/GraceNotes/GraceNotes/Features/Settings/Services/JournalDataExportService.swift
@@ -73,7 +73,8 @@ struct JournalDataExportService {
     }
 
     /// UTC `yyyyMMdd-HHmmss` stamp for export filenames; deterministic and locale-insensitive (`en_US_POSIX`).
-    /// Uses `Calendar` + `String(format:)` so concurrent exports (e.g. `Task.detached`) do not share a `DateFormatter`, which is not thread-safe.
+    /// Uses `Calendar` + `String(format:)` so concurrent exports (e.g. `Task.detached`) do not share a `DateFormatter`,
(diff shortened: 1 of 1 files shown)


kipyin:
We replaced the shared `DateFormatter` used for export filename timestamps with a formatter created per call on a UTC Gregorian calendar, so filenames stay stable and we avoid mutating formatter state across threads. Behavior stays aligned with `JournalDataExportServiceTests.test_exportFilenameTimestamp_matchesUTCGregorianPattern`. Per review, the doc comment is split across two lines so the line-length rule does not add SwiftLint noise. We did not add a concurrent stress test for `exportFilenameTimestamp`; existing coverage on the UTC/Gregorian pattern is enough for this change, and the `?? 0` fallback for missing components remains a theoretical edge case for normal dates.

PR #511, opened 2026-04-16
Harden Past journal search highlighting and stabilize match ordering

kipyin:
## Headline

Past search now avoids a pathological hang, keeps identical matches in a stable order, and uses a clearer debounce delay.

## User impact

Searching past journals should feel more predictable when multiple hits look the same, and the app avoids freezing if highlight logic ever runs with an empty query. The debounce behavior is unchanged in duration but expressed in a clearer, maintainable way.

## What changed

The debounced search wait now uses a duration-based sleep API instead of a raw nanosecond count, with the same quarter-second delay.

Within each day and section, search results that compare as equal under localized case-insensitive rules are ordered consistently by match id so the list does not jump unpredictably between updates.

The helper that finds highlight ranges in entry text now returns no ranges when the trimmed query is empty, so it cannot loop forever on a zero-length string match that never advances the search position.

## Verification

On a Mac with Xcode and the project’s grace CLI installed, run `grace ci` (or `grace test` with the usual simulator destination from gracenotes-dev.toml). Manually smoke-test Past search: type a query, confirm results and highlights update; try edge cases with whitespace-only input clearing to empty after trim if applicable.

## Risk

Medium

## Touch class

`ui-ux`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Views/PastJournalSearchViews.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Views/PastJournalSearchViews.swift
@@ -79,6 +79,7 @@ enum PastJournalSearchGrouping {
     }
 }
 
+/// File-scoped highlight rendering; ``text(content:highlightQuery:)`` is the only entry point used by views.
 private enum PastJournalSearchHighlighting {
     /// Body size matches ``AppTheme.warmPaperBody`` (17pt, scaled for Dynamic Type).
     private static let bodyPointSize: CGFloat = 17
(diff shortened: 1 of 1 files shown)


kipyin:
Thanks for the deep review and for calling out the empty-query risk: we added an early return in `matchRanges` so a blank query never drives `range(of:…)` into zero-width advances, which could otherwise spin on the main actor and freeze the UI if a future refactor dropped the outer empty check. We also tightened result ordering so ties sort in a stable, deterministic way instead of depending on incidental ordering. We agree tests for the empty-query case and for tie-breaking when localized comparison is equal would further lock this in; we can add those in a follow-up if you want that coverage on this PR. The lexical `id` tie-break is intentional for stability, with the tradeoff that it may not match “which duplicate came first” in the journal. Validation on our side is via `grace test` / `grace ci` on macOS with the iOS Simulator, as in the project docs.

PR #512, opened 2026-04-16
Harden ShareSheet popover configuration and async setup

kipyin:
## Headline

Sharing on iPad and large layouts should anchor the share popover reliably without odd presentation quirks.

## User impact

On iPad and other horizontally regular layouts, the share sheet uses a popover; misconfiguration can make sharing feel broken or visually wrong. This update lets the system choose sensible popover arrow directions and avoids touching a share controller after it may have been dismissed or torn down.

## What changed

The popover setup no longer clears permitted arrow directions, so the system can pick a natural anchor instead of forcing an empty set that can interact badly with layout. The deferred main-queue configuration now holds a weak reference to the activity view controller and bails out if it is gone, which prevents stray work on deallocated UI and reduces lifetime issues around the initial presentation frame.

## Verification

On a Mac with Xcode, build and run the GraceNotes scheme; on iPad simulator or a horizontally regular size class, open journal sharing and confirm the share popover presents with a stable anchor. Run `grace ci` (or `grace test` with the project’s default destination) to match CI.

## Risk

Medium

## Touch class

`ui-ux`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Views/ShareSheet.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Views/ShareSheet.swift
@@ -24,6 +24,8 @@ private enum ShareSheetPopover {
             width: 0,
             height: 0
         )
+        // Empty set avoids default arrow anchoring that can clash with chained sheets (e.g. journal share after another popover).
+        popover.permittedArrowDirections = []
     }
 }
 
--- a/GraceNotes/GraceNotes/Features/Journal/Views/ShareSheet.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Views/ShareSheet.swift
@@ -24,7 +24,6 @@ private enum ShareSheetPopover {
             width: 0,
             height: 0
         )
-        popover.permittedArrowDirections = []
     }
 }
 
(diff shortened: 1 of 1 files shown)


kipyin:
**Update:** We tightened popover setup by configuring the share activity controller on `viewDidLayoutSubviews` and `viewDidAppear`, plus a deferred pass on the main queue. The async closure uses `[weak controller]` and a `guard` so we do not keep the activity view controller alive after teardown; if that run is skipped, the layout/appearance hooks still apply `ShareSheetPopover.configureIfNeeded` when the view is valid. For horizontally regular layouts we set `permittedArrowDirections` to an empty set with a short comment beside it so the popover does not pick default arrows that can look wrong or fight chained presentations (for example journal share after another sheet)—matching the earlier behavior on iPad rather than UIKit defaults alone. Validation here is simulator- and Mac-only: please sanity-check journal share and settings export share on iPad or a regular-width layout when convenient; we did not add automated coverage for popover presentation. The existing `precondition(!activityItems.isEmpty)` contract is unchanged—callers should still avoid invoking the sheet with an empty item list.

PR #513, opened 2026-04-16
Fix completion info card when switching badges while it is open

kipyin:
## Headline

Switching to another completion badge no longer updates the card’s content before the dismiss animation finishes.

## User impact

If you opened the journal completion info card and tapped a different badge, the sheet could briefly show the new badge’s details while the old card was still closing, which felt glitchy and confusing. After this change, the selection updates when the card is ready to show the new information again, so the transition matches what you tapped.

## What changed

The logic for “card already open, user picks another badge” used to set the new badge as selected immediately, then run the close-and-reopen sequence. That meant the underlying selected badge could change during the exit animation or before the card came back. The handler now passes the intended badge into the scheduled close-and-reopen path and only updates the selected badge right before the entrance animation when that sequence runs. When the card was not already showing, behavior is unchanged: the selection is set, then the card is presented.

## Verification

On a Mac with Xcode and the iOS Simulator, run `grace ci` (or `grace test` per your usual workflow) and manually tap between completion badges with the info card open to confirm a clean transition with no wrong content during dismiss.

## Risk

Medium

## Touch class

`ui-ux`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Views/JournalCompletionInfoPresentation.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Views/JournalCompletionInfoPresentation.swift
@@ -25,12 +25,12 @@ final class JournalCompletionInfoPresentation {
             return
         }
 
-        selectedBadgeInfo = badgeInfo
-
         if isInfoCardPresented {
-            scheduleInfoCardCloseThenReopenAfterDelay(reduceMotion: reduceMotion)
+            scheduleInfoCardCloseThenReopenAfterDelay(newBadgeInfo: badgeInfo, reduceMotion: reduceMotion)
             return
         }
+
+        selectedBadgeInfo = badgeInfo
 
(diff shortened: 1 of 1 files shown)


kipyin:
This branch already sets `selectedBadgeInfo` synchronously at the start of `scheduleInfoCardCloseThenReopenAfterDelay` (right after kicking off the exit animation), so inline pills pick up the new tier tint immediately during the close-and-reopen sequence instead of waiting for the delayed task. The earlier bug was that the reopen path did not carry the tapped badge through correctly and could update selection at the wrong time relative to the first-open path; fixing that ordering keeps `selectedBadgeInfo` in sync with intent without flashing the new card content during removal. I did not add automated coverage for rapid badge switches versus dismiss versus `cancelTasksOnDisappear` interleaving; that remains a good follow-up. I will run a quick manual pass on a Mac with Reduce Motion on and off (and normal timing) to confirm the card and toolbar feel right. The sticky toolbar chip still reflecting the current journal tier rather than a “pending” tier when completion changes mid-sequence is a known edge case that predates this change.

PR #515, opened 2026-04-16
Align Sprout unlock suppression with real App Tour eligibility

kipyin:
## Headline

Sprout feedback no longer hides during UI tests when the tour itself cannot appear.

## User impact

When the App Tour is blocked (for example under UI tests), the Sprout-stage unlock toast and header celebration were still being suppressed as if the full-screen tour were about to show. That could hide expected feedback in tests and misrepresent first-session behavior in those runs. The policy now only suppresses that feedback when the tour would actually be eligible to present.

## What changed

The sprout-unlock suppression helper now takes the same structured orientation inputs used elsewhere instead of separate boolean parameters, so call sites can pass through `todayOrientationInputs()` without splitting fields. A UI-test guard was added so suppression matches `appTourOutcome` and `AppTourTrigger.evaluate`: if the tour is not going to present (including because UI tests suppress it), the sprout toast and celebration are not withheld for that reason. Documentation was updated to describe this alignment and to note that guided-journal completion on the inputs struct is carried for convenience at the call site but is not used by this particular decision.

## Verification

On a Mac with Xcode and the project’s `grace` CLI, run `grace ci` (or `grace test` with your usual simulator destination) to confirm the GraceNotes scheme builds and tests pass.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Tutorial/JournalTodayOrientationPolicy.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Tutorial/JournalTodayOrientationPolicy.swift
@@ -55,4 +55,30 @@ enum JournalTodayOrientationPolicy {
         guard newLevel == .sprout else { return false }
         return appTourOutcome(for: orientationInputs) != nil
     }
+
+    /// Bridge for call sites that pass orientation fields separately (for example ``JournalScreen``), using
+    /// the same app-wide defaults for UITesting and guided-journal completion as a fully specified ``Inputs``.
+    static func shouldSuppressSproutUnlockToast(
+        isTodayEntry: Bool,
+        newLevel: JournalCompletionLevel,
+        hasSeenAppTour: Bool,
+        milestoneHighlight: JournalUnlockMilestoneHighlight,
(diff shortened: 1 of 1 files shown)


kipyin:

kipyin:
Thanks for the detailed review. **What changed:** Sprout unlock toast suppression for Today now uses the same eligibility path as the App Tour by delegating to `appTourOutcome(for:)` (after the existing sprout / milestone guards), so UI-test gating and `AppTourTrigger.evaluate` cannot drift from the suppression logic. **Call sites:** `JournalScreen` keeps using the convenience overload that builds `Inputs` with the same orientation fields plus `ProcessInfo.graceNotesIsRunningUITests` and guided-journal completion from `UserDefaults`, matching production behavior without duplicating preconditions at the call site. **Tests:** `JournalTodayOrientationPolicyTests` were updated to use the current API so the test target builds, and existing `appTourOutcome` cases already cover “UI tests ⇒ no tour outcome.” I will run `grace ci` on macOS to confirm the full simulator build and `GraceNotesTests` as noted in `AGENTS.md`.

PR #516, opened 2026-04-16
Speed up streak calculation with a single journal pass

kipyin:
## Headline

Streak counts now build both completion maps in one scan and reuse yesterday’s date.

## User impact

Streak numbers and “done today” behavior stay the same, but computing them does less work. That keeps streak updates snappier when you have many journal days loaded and avoids repeating the same date math for basic and perfect streaks.

## What changed

Streak summary used to walk every journal entry twice—once to see which days had meaningful content and again for “perfect” (full harvest) days—because two separate helpers each built a per-day map. That is now a single loop that fills both maps at once. The calendar step that finds “yesterday” relative to today is done once in the summary and passed into streak length, instead of being recalculated inside each streak path. The private helper that only built one map was removed because its job is folded into that loop.

## Verification

On a Mac with Xcode and the project’s dev CLI installed, run `grace ci` (or `grace test` with your usual simulator destination) so lint and simulator tests cover this path.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Services/StreakCalculator.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Services/StreakCalculator.swift
@@ -23,33 +23,25 @@ struct StreakCalculator {
 
     func summary(from entries: [Journal], now: Date = .now) -> StreakSummary {
         let today = calendar.startOfDay(for: now)
-        let basicByDay = buildCompletionByDay(entries: entries) { $0.hasMeaningfulContent }
-        // "Perfect" = Harvest: all fifteen chips. `completedAt` alone must not inflate this streak.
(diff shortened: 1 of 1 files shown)


kipyin:
The file already includes the restored comment on the `perfectByDay` merge (lines 32–33). Here is the comment body only:

---

**Update:** Restored a short inline note on the `perfectByDay` merge so it stays obvious that the perfect streak is driven by harvest / Bloom (`hasReachedBloom`), not `completedAt`, and that completion timestamps must not inflate that streak. That matches the previous intent and guards against a future “optimization” that lines perfect streak up with completion time. No behavior change beyond the single-pass refactor already in this PR. Please run `grace test` (or Xcode) on macOS to confirm streak tests if you want runtime verification—CI here can’t execute `GraceNotesTests`.

PR #517, opened 2026-04-16
Harden export history detail normalization for whitespace and invisible text

kipyin:
## Headline

Backup export history hides meaningless detail strings and reads more cleanly when spacing is messy.

## User impact

Export history lines and VoiceOver labels no longer surface odd gaps from extra spaces or tabs, and they omit detail text that would only show invisible characters (for example zero-width format characters). That keeps Settings clearer and avoids confusing or empty-looking metadata.

## What changed

The shared normalizer for export history detail now collapses consecutive horizontal whitespace to a single space after line breaks are flattened. It also checks Unicode scalar categories so control and format characters do not count as visible content; if nothing visible remains, the detail is treated as absent, matching the intent that only meaningful text appears in the row and accessibility strings.

## Verification

On a Mac with the repo and Xcode, run `grace ci` (or `grace test` per your usual workflow) to confirm the project still builds and tests pass; optionally spot-check Settings import/export history with odd spacing or edge-case detail strings if you have fixtures.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Settings/ImportExportTechnicalDetailFormatting.swift
+++ b/GraceNotes/GraceNotes/Features/Settings/ImportExportTechnicalDetailFormatting.swift
@@ -4,6 +4,7 @@ struct ExportHistoryLineParts {
     let kindLabel: String
     let statusLabel: String
     /// Non-nil when `entry.detail` has visible content after normalization (trim; line breaks flattened).
+    /// For this purpose, Unicode scalars in General Category `.control` or `.format` count as non-visible.
     let detail: String?
 }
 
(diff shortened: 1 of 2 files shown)


kipyin:
This worktree already includes aligned documentation (export detail visibility ignores **all** Unicode General Category `.control` and `.format` scalars, with examples in the doc comment) and `test_exportHistoryLineParts_formatAndControlOnlyDetailIsNil`, which asserts format/control-only detail becomes `nil` and the plain label has only two segments. Here is the PR comment body only:

---

Thanks for the deep review. I updated the documentation so it matches the implementation: “has visible content” means there is at least one scalar that is **not** in General Category `.control` or `.format` (including zero-width, joiners, and BOM-style scalars), so nobody mistakes this for a narrow zero-width-only rule. I added `test_exportHistoryLineParts_formatAndControlOnlyDetailIsNil` with a detail string composed only of `.format`/`.control` scalars and assert `parts.detail` is `nil` and the plain label has exactly two “·”-separated segments. Full compile and test coverage for this target still needs to run on macOS via `grace ci` / Xcode; behavior here remains best-effort for single-line display, not a full grapheme audit, and normalization is presentation-only.

PR #518, opened 2026-04-16
Use shared export filename timestamp for scheduled backups

kipyin:
## Headline

Scheduled folder backups now use the same timestamp rules as the rest of journal export naming.

## User impact

Backup file names stay aligned with exports you trigger elsewhere in the app, so files sort and read consistently in the chosen folder. One shared implementation also means future timestamp tweaks cannot silently diverge between flows.

## What changed

Scheduled backup filenames no longer configure their own DateFormatter for the time stamp. The runner asks JournalDataExportService for the stamp via exportFilenameTimestamp(for:) and keeps the existing grace-notes-scheduled prefix, UUID suffix, and .json extension. Behavior stays the same idea—sortable, unique names—while centralizing how the stamp is produced.

## Verification

On macOS with Xcode and the iOS Simulator available, run grace ci from the repository root (or grace test with your usual destination from gracenotes-dev.toml) to confirm the GraceNotes scheme builds and automated checks pass.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Settings/Services/ScheduledBackupRunner.swift
+++ b/GraceNotes/GraceNotes/Features/Settings/Services/ScheduledBackupRunner.swift
@@ -122,13 +122,9 @@ enum ScheduledBackupRunner {
     }
 
     private static func copyTempExport(at tempFile: URL, to folderURL: URL, exportDate: Date) throws -> String {
-        let formatter = DateFormatter()
-        formatter.calendar = Calendar(identifier: .gregorian)
-        formatter.locale = Locale(identifier: "en_US_POSIX")
-        formatter.timeZone = TimeZone(secondsFromGMT: 0)
-        formatter.dateFormat = "yyyyMMdd-HHmmss"
+        let stamp = JournalDataExportService().exportFilenameTimestamp(for: exportDate)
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

Thanks for the thorough pass—agreed there are no blocking issues.

This change removes a duplicated `DateFormatter` in the temp-export path and routes the timestamp through `JournalDataExportService.exportFilenameTimestamp(for:)`, so scheduled backups use the same UTC Gregorian `en_US_POSIX` `yyyyMMdd-HHmmss` pattern that `JournalDataExportServiceTests` already covers. `performScheduledExport` now derives a single `exportDate` and passes it to both `exportArchiveFile(..., now:)` and the destination filename, which keeps the archive prefix and final backup name aligned for that instant. Existing `copyTempFile` behavior (remove-then-copy, UUID suffix) and pruning/retention are unchanged.

On the follow-ups: a small unit test tying the on-disk filename stamp to `exportedAt` in the JSON for a fixed clock would be a good lock-in; the static formatter’s thread-safety note matches pre-existing usage and is worth keeping in mind if exports ever fan out concurrently; full security-scoped backup flow remains validated by the macOS simulator CI path rather than Linux.

PR #519, opened 2026-04-16
Normalize weekly insight theme keys when skipping duplicates

kipyin:
## Headline

Weekly review insights now treat themes as the same using the app’s theme normalization, not just lowercase text.

## User impact

You are less likely to see two very similar weekly insights about the same underlying theme when one should be hidden as a duplicate. This makes the review feel more coherent and avoids repetitive prompts.

## What changed

The duplicate check for selected weekly insights compared primary themes with a simple lowercased string. That could miss matches that the rest of journaling already treats as the same theme, or treat messy empty strings inconsistently. The logic now uses the shared text normalizer’s theme label normalization for both the candidate and already-selected insights, and it skips comparisons when normalization yields an empty string. Full-completion patterns are still excluded from this skip rule as before.

## Verification

On a Mac with Xcode, run `grace ci` (or `grace test` with the project’s default simulator destination) to build and run automated checks for the GraceNotes scheme.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Services/WeeklyInsightCandidateBuilder+Candidates.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Services/WeeklyInsightCandidateBuilder+Candidates.swift
@@ -257,18 +257,27 @@ extension WeeklyInsightCandidateBuilder {
         }
     }
 
+    /// Key used to detect duplicate themes when selecting weekly insights. Prefer normalized labels;
+    /// when normalization is empty (whitespace-only, punctuation-only, etc.), fall back to trimmed
+    /// lowercased raw text so meaningless themes still dedupe like the pre-normalization path.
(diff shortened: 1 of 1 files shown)


kipyin:
Thanks for the detailed review and for flagging the empty-normalization edge case.

We added a small `themeDedupeKeyForSkip` helper so weekly insight selection still uses normalized theme labels when they are meaningful, but when normalization strips down to nothing (whitespace-only, punctuation-only, etc.) we fall back to trimmed lowercased raw text. That restores the old behavior where two “empty” or meaningless themes could still match for deduplication instead of skipping the dedupe path entirely. No other product semantics were changed on purpose.

We have not yet added unit tests that lock in punctuation, diacritics, and case-only differences for `shouldSkip` / `selectInsights`; that remains a reasonable follow-up if we want stronger regression coverage.

PR #522, opened 2026-04-16
Treat cancellation-like errors as benign when reading iCloud account status

kipyin:
## Headline

Stop flagging normal CloudKit or URL cancellations as iCloud status failures in logs.

## User impact

When iCloud work is cancelled or interrupted in ways CloudKit reports as operation cancelled (not only Swift task cancellation), Settings and diagnostics see fewer scary “failed to fetch iCloud account status” messages. That makes real sync or account problems easier to spot and reduces noise during normal navigation or teardown.

## What changed

The iCloud account status fetch already had a helper that recognizes several cancellation shapes (Swift cancellation, CloudKit operation cancelled, URL cancelled, and nested underlying errors). The catch path only skipped logging for plain Swift CancellationError, so other cancellation forms were still logged as errors even though the code already treated any failure the same for the user-facing bucket. The catch path now uses the same cancellation detection before logging, so benign cancellations are not reported as failures.

## Verification

On a Mac with Xcode and the repo’s grace CLI installed, run `grace ci` (or `grace test` with the project’s default simulator destination) to confirm the GraceNotes scheme still builds and tests pass.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Settings/Services/ICloudAccountStatusService.swift
+++ b/GraceNotes/GraceNotes/Features/Settings/Services/ICloudAccountStatusService.swift
@@ -23,7 +23,7 @@ final class ICloudAccountStatusService: ICloudAccountStatusProviding {
             }.value
             return ICloudAccountBucket(status)
         } catch {
-            if !(error is CancellationError) {
+            if !isCancellationLike(error) {
                 let nsError = error as NSError
                 iCloudAccountStatusLogger.error(
                     """


kipyin:
**Sentry — review feedback**

Thanks for the detailed review—glad the `isCancellationLike` gate reads as the right match for CloudKit/URLSession-style cancellations and for superseding `refresh()` on `ICloudAccountStatusModel`.

We only broadened the **logging** path in `fetchAccountBucket()` so benign cancellations are treated the same way as elsewhere in this flow; the outcome is still **`.couldNotDetermine`**, so we do not mis-report iCloud availability when a fetch is cancelled or torn down.

I agree on the residual risk: if Apple adds another cancellation shape, we might still log noise until we extend `isCancellationLike`. Full reassurance on rapid refresh/navigation still belongs on **macOS + Xcode / Simulator**, consistent with how this repo validates (`grace test` / `grace ci`).

PR #523, opened 2026-04-16
Remove duplicate reminder time parsing from daily notification sync

kipyin:
## Headline

Daily reminder rescheduling now relies on one shared, validated clock-time reader.

## User impact

Reminder scheduling behavior stays the same for people who use daily reminders, but the app no longer carries two different ways to interpret the saved time from UserDefaults. That reduces the risk of future edits accidentally changing behavior in one path but not the other.

## What changed

`DailyReminderNotificationSync` already used `ReminderSettings.coercedTimeInterval(fromUserDefaults:)` when computing the next reminder time. A private helper that duplicated the same UserDefaults parsing and validation was left in the file but never called, so it was removed. Reminder time coercion and non-finite value handling remain centralized in `ReminderSettings`, which is the single place reviewers should look for that logic.

## Verification

On a Mac with Xcode and the project’s `grace` CLI installed, run `grace ci` (or `grace test` with the repo’s default simulator destination) to confirm the GraceNotes scheme still builds and tests pass.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Services/Reminders/DailyReminderNotificationSync.swift
+++ b/GraceNotes/GraceNotes/Services/Reminders/DailyReminderNotificationSync.swift
@@ -27,25 +27,4 @@ enum DailyReminderNotificationSync {
         }
         _ = await reminderScheduler.rescheduleEnabledReminder(at: reminderTime, body: body)
     }
-
-    /// Reads the saved clock time. `UserDefaults` is untyped; accept `NSNumber` and reject non-finite values.
-    private static func storedReminderTimeInterval(from userDefaults: UserDefaults) -> TimeInterval {
-        guard let raw = userDefaults.object(forKey: ReminderSettings.timeIntervalKey) else {
-            return ReminderSettings.defaultTimeInterval
-        }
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

Removed the unused `storedReminderTimeInterval(from:)` helper from daily reminder sync code. Nothing called it; rescheduling already reads and sanitizes the stored interval only through `ReminderSettings.coercedTimeInterval(fromUserDefaults:)` and related APIs, so this was duplicate, dead parsing surface and deleting it does not change runtime behavior. Keeping a single reader in `ReminderSettings` matches what the review called out and avoids two different interpretations of the same UserDefaults key. On validation: `grace test` / `grace ci` on macOS still apply for automated checks; a quick manual pass on simulator or device that the daily reminder still fires at the set time after editing a journal remains the practical guard for reschedule behavior, as the bot noted. If we ever see odd values from old storage shapes, we can add a focused unit test on `ReminderSettings.coercedTimeInterval` rather than reintroducing another reader here.

PR #524, opened 2026-04-16
Stabilize most recurring theme rows and accessibility keys

kipyin:
## Headline

Separates SwiftUI row identity from UI-test identifiers so recurring-theme rows update reliably when labels collide.

## User impact

When two recurring themes looked the same by label, list updates could behave oddly and automation could target the wrong row. The card now keys rows in a way that matches the theme value while still using a composite key for accessibility identifiers when labels are not unique.

## What changed

The most recurring themes list no longer uses a custom string property as the ForEach identity. Rows are keyed by the theme value itself, which is appropriate when the model differentiates themes beyond the display label. The old composite string used for stable differentiation was moved into a small helper used only for accessibility and UI-test identifiers, so test IDs stay stable where labels can collide without tying that logic to list animation and updates.

## Verification

On a Mac with Xcode, run `grace ci` (or `grace test` with the project’s default destination) to lint and build. Manually open Review insights with overlapping or similarly labeled recurring themes and confirm rows update; if you have UI tests that hit `MostRecurringThemeRow.*` identifiers, re-run those to confirm they still resolve distinct rows.

## Risk

Medium

## Touch class

`ui-ux`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Views/ReviewMostRecurringCard.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Views/ReviewMostRecurringCard.swift
@@ -82,7 +82,7 @@ struct ReviewMostRecurringCard: View {
         ) {
             VStack(alignment: .leading, spacing: 10) {
                 VStack(alignment: .leading, spacing: 8) {
-                    ForEach(Array(themes.prefix(3)), id: \.mostRecurringRowIdentity) { theme in
+                    ForEach(Array(themes.prefix(3)), id: \.self) { theme in
                         mostRecurringThemeRow(theme)
                     }
                 }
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

Thanks for the detailed review—agreed there are no blocking issues.

We switched the recurring-theme rows to `ForEach(..., id: \.self)` so each row’s identity follows the full `ReviewMostRecurringTheme` value (synthesized `Hashable`), which matches the “one row per canonical key” pipeline and avoids treating two different themes as the same row when they share a display label. Accessibility identifiers still use `mostRecurringAccessibilityKey`, aligned with the old `mostRecurringRowIdentity` inputs (label, counts, sorted evidence ids), so UI tests and VoiceOver should see stable ids for the same underlying row.

On residual risk: if the pipeline ever produced duplicate keys in the list, SwiftUI could merge rows—worth a quick `grace test` / Xcode pass on the Journal review UI if that code path changes. I’m not planning extra UITest assertions for same-label rows unless we see that scenario in the wild; the sanitizer collision case remains a theoretical tail risk given how composite keys are built.

Full simulator verification should be run on a Mac per project workflow (`grace test` / `grace ci`); this environment only ran `swiftlint lint`.

PR #525, opened 2026-04-16
Use monotonic system uptime for performance trace timing

kipyin:
## Headline

Performance logs now measure durations with a steady clock, so timings stay trustworthy when the clock jumps.

## User impact

Developers and diagnostics that rely on performance traces get durations that reflect real elapsed work instead of misleading spikes or negatives if the device time changes. This makes debugging and profiling notes from the field more reliable.

## What changed

Performance trace begin/end no longer uses wall-clock time from Core Foundation. It uses the system uptime clock instead, which moves forward in a stable way and is suited to measuring intervals. Elapsed milliseconds are clamped so logged durations never read as negative if something odd happens at the boundary.

## Verification

On a Mac with the repo and Xcode, run `grace ci` (or `grace test` with the project’s usual simulator destination) to confirm the app still builds and tests pass; optionally trigger a code path that logs performance traces and confirm end lines show sensible millisecond values.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Application/PerformanceTrace.swift
+++ b/GraceNotes/GraceNotes/Application/PerformanceTrace.swift
@@ -9,11 +9,11 @@ enum PerformanceTrace {
 
     static func begin(_ label: String) -> TimeInterval {
         logger.log("[BEGIN] \(label, privacy: .public)")
-        return CFAbsoluteTimeGetCurrent()
+        return ProcessInfo.processInfo.systemUptime
     }
 
     static func end(_ label: String, startedAt start: TimeInterval) {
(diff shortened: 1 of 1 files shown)


kipyin:
We now detect when `systemUptime` goes backward relative to the stored `begin` token and treat that as an invalid span instead of reporting **0 ms**. In that case we log `[END]` with an explicit “invalid span; monotonic clock decreased” message and skip emitting a duration, so triage is not misled by a fake “instant” measurement. In **DEBUG** builds we also `assert` to catch misuse early. This keeps duration traces on monotonic uptime for NTP safety while making exceptional clock discontinuities or future cross-lifetime token mistakes visible in logs rather than hidden behind a clamp.

PR #526, opened 2026-04-16
Keep scheduled backup single-flight until MainActor bookkeeping finishes

kipyin:
## Headline

Scheduled backups no longer release their overlap lock before last-run time and history are updated.

## User impact

Rapid app activations could start a second scheduled backup while the first run was still finishing preference and export-history updates on the main thread, risking duplicate work or confusing backup status. The lock now covers the full run so only one scheduled backup flow is in flight until everything is recorded.

## What changed

The single-flight guard for scheduled backups used to end as soon as the background export task returned, which was before updating last-run time and export history on the main actor. The completion handler was moved so the guard stays active for the entire async function, including main-thread bookkeeping. The documentation was updated to describe that behavior and why each activation can spawn a new task.

## Verification

On a Mac with Xcode and the project’s grace CLI installed, run `grace ci` from the repo root (or `grace test` with the usual simulator destination) to lint and build; exercise scheduled backup by switching away and back quickly with backups enabled and confirm only one run completes before the next can start.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Settings/Services/ScheduledBackupRunner.swift
+++ b/GraceNotes/GraceNotes/Features/Settings/Services/ScheduledBackupRunner.swift
@@ -7,8 +7,11 @@ private let scheduledBackupLogger = Logger(
     category: "ScheduledBackup"
 )
 
-/// Prevents overlapping scheduled backup runs until `runIfDue` finishes, including MainActor
-/// bookkeeping (`lastRunAt`, export history). Each `active` transition spawns a new `Task` in ``GraceNotesApp``.
+/// Prevents overlapping scheduled backup runs until the detached export + MainActor bookkeeping
+/// completes. `end()` runs in a `defer` on that detached task so `inFlight` is not cleared if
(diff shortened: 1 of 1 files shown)


kipyin:
**Response to review**

We moved `scheduledBackupSingleFlight.end()` into a `defer` at the start of the detached `Task` body so the single-flight gate stays set until that detached work finishes (export, MainActor bookkeeping, and failure recording), not when the outer `await` returns. That fixes the overlap we saw when the scene became active again before `lastRunAt` and history were updated on the main actor. It also addresses the cancellation concern you raised: if whatever awaits `runIfDue` is cancelled while suspended on the detached task’s result, the waiter can unwind without clearing `inFlight` early, because `end()` only runs when the detached closure exits. The doc comment on `ScheduledBackupSingleFlight` states that `GraceNotesApp` uses a new `Task` per `active` transition rather than a long-lived cancellable parent. We agree rapid `scenePhase` churn is still hard to lock in with a unit test; we’ll keep an eye on `.public` logging if error descriptions ever start carrying sensitive paths.

PR #527, opened 2026-04-17
Fix inline sentence editor caret jumps and Dynamic Type minimum height

kipyin:
## Headline

Journal sentence fields keep a stable caret while typing and size correctly with body text settings.

## User impact

Editing gratitudes and other inline sentences should feel steadier: the insertion point and selection no longer jump when SwiftUI refreshes the field while you are typing. Minimum height now follows the scaled body font, so the editor layout stays sensible when you change text size or Dynamic Type.

## What changed

The UITextView bridge used to apply binding updates by assigning the full string whenever it differed from the view, which resets the selection while the field is first responder. The view now reapplies external text with the caret restored and clamped to valid UTF-16 ranges so edits stay where you expect.

The minimum height was a static value computed once from the body font line height. It is now computed on demand so it tracks UIFontMetrics scaling and stays consistent with the live font used for layout and measurement.

## Verification

On a Mac with Xcode and the iOS Simulator, run `grace ci` (or `grace test` with the project’s default destination). Manually open Today journal editing, type in an inline sentence field with the keyboard up, and change Settings ▸ Accessibility ▸ Display & Text Size ▸ Larger Text while the field is visible to confirm layout and caret behavior.

## Risk

Medium

## Touch class

`ui-ux`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Views/InlineSentenceEditorField.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Views/InlineSentenceEditorField.swift
@@ -12,10 +12,20 @@ private enum InlineSentenceEditorFieldLayout {
     static let bodyFontName = "SourceSerif4Roman-Regular"
     static let bodyFontSize: CGFloat = 17
 
+    private static var cachedScaledBodyFont: UIFont?
+    private static var cachedContentSizeCategoryForFont: UIContentSizeCategory?
+
     static func bodyUIFont() -> UIFont {
+        let category = UITraitCollection.current.preferredContentSizeCategory
+        if let cached = cachedScaledBodyFont, cachedContentSizeCategoryForFont == category {
+            return cached
+        }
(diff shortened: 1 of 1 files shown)


kipyin:
Thanks for the deep review—agree there are no blockers.

This updates the inline sentence editor’s `UITextView` bridge so external `text` binding sync does not overwrite text while IME composition is active (`markedTextRange`), and when we do sync while focused we restore the selection so the caret does not jump. We also stop a focus loop where UIKit could end editing before SwiftUI cleared `@FocusState`, which was causing repeated `becomeFirstResponder` churn and keyboard noise. Layout uses a cached Dynamic Type–scaled body font, a minimum height that matches that font, and a capped multiline height with scrolling for very long content. Persistence and export paths are unchanged; validation is `grace ci` on Mac plus manual checks for inline editing, Larger Text, and Japanese/Chinese keyboards as you noted.

PR #528, opened 2026-04-17
Sync reminder toggle state after enabling or disabling daily reminders

kipyin:
## Headline

Daily reminder switch in Settings now stays aligned with the real notification state after you change it.

## User impact

Turning daily reminders on or off could leave the toggle out of step with what the app actually scheduled or what iOS allowed. The switch and related UI now update from the live reminder status right after that change finishes, so what you see matches whether reminders are really on.

## What changed

The Settings screen already kept the reminder toggle in sync when the underlying live status changed from elsewhere. After the async work to turn reminders on or off completes, the screen now runs the same sync step it uses on load and on status changes, so the local toggle and guidance reflect the updated permission and scheduling outcome immediately.

This is a small behavioral fix in the reminders section only; no new strings or layout changes.

## Verification

On a Mac with Xcode and the project’s grace CLI installed, run `grace ci` (or `grace test` with your usual simulator destination) from the repo root to lint and build. Manually: open Settings, toggle Daily reminders on and off (including after denying or restoring notification permission) and confirm the switch and messaging stay consistent with the actual state.

## Risk

Medium

## Touch class

`ui-ux`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Settings/SettingsScreen.swift
+++ b/GraceNotes/GraceNotes/Features/Settings/SettingsScreen.swift
@@ -205,6 +205,7 @@ private extension SettingsScreen {
                 isReminderPickerExpanded = newValue
                 Task {
                     await reminderState.setReminderEnabled(newValue)
+                    syncReminderControlState(with: reminderState.liveStatus)
                 }
             }
         )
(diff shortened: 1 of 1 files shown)


kipyin:
**What changed**

After `setReminderEnabled(false)` returns, we no longer always re-sync the toggle from `liveStatus`. If the user turned reminders off but `liveStatus` is still `.enabled`—because `disableReminders()` deferred work while another reminder operation was in progress—we skip `syncReminderControlState` so the optimistic OFF is not flipped back on until `liveStatus` actually updates. Existing `.onChange(of: reminderState.liveStatus)` still reconciles the control when the deferred disable finishes. This addresses the regression where a post-await sync could undo the user’s choice until the queue drained.

PR #530, opened 2026-04-17
Avoid allocating dates array when finding earliest journal day

kipyin:
## Headline

Weekly review rhythm stats compute the earliest entry day without materializing a full day array.

## User impact

Users with large journals get snappier weekly review and past-statistics aggregation with less transient memory use while scrolling or refreshing review screens. Behavior stays the same: the code still finds the minimum calendar day the same way, just without building every intermediate date upfront.

## What changed

Computing the earliest calendar day across all entries used a non-lazy map, which built a full array of normalized day values before taking the minimum. That extra allocation scales with total entries and is unnecessary for a min scan. The path now uses a lazy map so only one day at a time is produced while the minimum is computed, cutting peak allocation for the same logical result.

## Verification

On a Mac with Xcode, run `swiftlint lint` on the touched file or repo, then `grace ci` (or `grace test`) from the project root to confirm the GraceNotes scheme still builds and tests pass.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Services/WeeklyReviewAggregates.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Services/WeeklyReviewAggregates.swift
@@ -409,7 +409,7 @@ private extension WeeklyReviewAggregatesBuilder {
         let weekLastStart = calendar.startOfDay(for: weekLastInclusive)
         let referenceDayStart = calendar.startOfDay(for: referenceDate)
         let endDayInclusive = min(weekLastStart, referenceDayStart)
-        guard let entryMinRaw = allEntries.map({ calendar.startOfDay(for: $0.entryDate) }).min() else {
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

Thanks for the thorough review. This change updates `buildRhythmHistory` so we derive the earliest journal calendar day with a single lazy pass over entry dates instead of materializing an intermediate `[Date]` before taking the minimum, which trims allocation while keeping the same lower-bound logic (Past window clamp, rhythm horizon cap, and dense day rows). The work is scoped to that rhythm-history path—not a repo-wide pass over every “earliest day” pattern—so the PR title points at the behavioral anchor we touched rather than implying a sweeping refactor. I agree large journals still pay O(n) `startOfDay` work per build; that tradeoff matches the goal here. Validation is best done on macOS with `grace test` / `grace ci` and a quick simulator check of Past / weekly review rhythm UI if you want extra confidence on a path that is not exercised every day.

PR #531, opened 2026-04-17
Handle non-string UserDefaults values in journal appearance migration

kipyin:
## Headline

Today tab journal appearance now migrates and normalizes even when the saved preference is not a string.

## User impact

If the journal appearance key was ever stored as a non-string (or otherwise read as nil by `string(forKey:)`), the app could skip cleanup and leave `@AppStorage` out of sync with what you see. After this change, those cases are treated as empty, resolved to the correct mode, and rewritten to a canonical string so the Today tab stays predictable and settings stay consistent.

## What changed

The migration helper for journal appearance no longer exits early whenever `UserDefaults.string(forKey:)` is nil. It now checks whether anything is stored under the key at all: real strings are used as before; if a value exists but is not a string, it is treated as an empty string so the existing normalization path can pick a valid mode and persist the canonical raw value. If the key is absent entirely, the function still returns without writing.

## Verification

On a Mac with Xcode and the project’s `grace` CLI, run `grace ci` (or `grace test` per your usual profile) to confirm lint and simulator builds pass. Optionally reset or seed `UserDefaults` with a non-string for `journalTodayAppearanceMode`, launch the app, and confirm Today appearance and migration behave as expected.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/DesignSystem/JournalAppearanceMode.swift
+++ b/GraceNotes/GraceNotes/DesignSystem/JournalAppearanceMode.swift
@@ -23,7 +23,15 @@ enum JournalAppearanceMode: String, CaseIterable, Identifiable, Sendable {
     /// surrounding whitespace, unknown values, etc.).
     static func migrateLegacyJournalAppearanceRawValueIfNeeded(defaults: UserDefaults = .standard) {
         let key = JournalAppearanceStorageKeys.todayMode
-        guard let stored = defaults.string(forKey: key) else { return }
+        let stored: String
+        if let string = defaults.string(forKey: key) {
+            stored = string
+        } else if defaults.object(forKey: key) != nil {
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

Addressed the suite teardown feedback by clearing each isolated `UserDefaults` suite with `removePersistentDomain(forName:)` on the same `UserDefaults(suiteName:)` instance used in the test, so teardown reliably wipes the suite and stays consistent with other tests in the repo. Added a focused migration test that stores a non-string value under `journalTodayAppearanceMode`, runs `migrateLegacyJournalAppearanceRawValueIfNeeded`, and asserts the key is rewritten to the canonical `"standard"` string so CI covers the `object(forKey:) != nil` / `string(forKey:) == nil` branch. Verified with `swiftlint` on the updated test file; please run `grace test` on a Mac to confirm the suite passes in the simulator.

PR #532, opened 2026-04-17
Fix rhythm strip pinning by skipping list/table scroll views

kipyin:
## Headline

The Past tab rhythm strip pins to its horizontal chart scroll, not the surrounding list.

## User impact

End-of-strip pinning and related scroll observation could latch onto the wrong scroll surface when the rhythm row sits inside a list, because the nearest ancestor UIScrollView is often the list or table host, not the horizontal strip. That could make the pin track vertical list scrolling or other containers instead of the rhythm chart. Users should see stable horizontal rhythm behavior while the list still scrolls normally.

## What changed

Scroll-view discovery for the rhythm strip no longer returns the first UIScrollView found walking up the view hierarchy. It now skips UITableView, UICollectionView, and UITextView so we resolve the embedded horizontal rhythm ScrollView instead of list, collection, or text scroll hosts (which are all UIScrollView subclasses). The helper was renamed and documented to state that intent, and the attach path uses the new lookup.

## Verification

On macOS with Xcode and the iOS Simulator, run `grace ci` (or `grace test` with the project’s default destination from `gracenotes-dev.toml`). In the app, open the Past tab, use the “Days you wrote” rhythm strip inside the list, and scroll the strip horizontally; confirm end-pinning and edge treatment follow the horizontal strip while vertical list scrolling still behaves as expected.

## Risk

Medium

## Touch class

`ui-ux`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Views/ReviewSummaryCard.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Views/ReviewSummaryCard.swift
@@ -571,7 +571,7 @@ struct ReviewDaysYouWrotePanel: View {
             columnWidth = (64 * scale).rounded(.toNearestOrAwayFromZero)
             columnGap = max(2, (3 * scale).rounded(.toNearestOrAwayFromZero))
             columnCornerRadius = (10 * scale).rounded(.toNearestOrAwayFromZero)
-            columnEdgeInset = max(2, (3 * scale).rounded(.toNearestOrAwayFromZero))
+            columnEdgeInset = max(2, (3 * scale).rounded(.toNearestOrAwayFromZero)
             rowHeight = (34 * scale).rounded(.toNearestOrAwayFromZero)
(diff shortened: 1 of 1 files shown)


kipyin:
Here’s a concise PR comment you can paste:

---

We narrowed the coordinator’s ancestor walk so it skips only the list row host (`UITableView`) and `UITextView`, and no longer treats every `UICollectionView` as something to ignore. That keeps trailing-edge pinning attached to the rhythm strip’s own horizontal scroller today and avoids mis-pinning if the strip is ever implemented with a collection view. On the UI test side, the Past-tab wide-rhythm tests now wait for the `ReviewRhythmHorizontalScroll` scroll view (the same identifier the rhythm `ScrollView` exposes under UI tests) before asserting that the newest day column is on-screen and the oldest is off-screen in portrait. Together, that gives automated coverage that the horizontal strip we care about is present in the real list/host context and stays pinned to the trailing edge after layout for wide seeded rhythm content.

PR #534, opened 2026-04-17
Use POSIX locale for journal appearance key normalization

kipyin:
## Headline

Journal appearance settings now match stored values consistently regardless of device language.

## User impact

If you had changed the Today journal look in Settings, the app could misread the saved value in rare cases where lowercasing depends on language or region (for example, Turkish locale and certain letters). The journal chrome should now stay on Bloom or Standard as intended after relaunch or migration.

## What changed

The change normalizes saved appearance strings with a fixed English POSIX locale instead of the user’s current locale, so trimming and lowercasing behave like stable identifier matching rather than prose. The legacy Summer-to-Bloom string is a named constant, and the documentation now states that recognizable values still map to Bloom or Standard while unknown values fall back to Standard.

## Verification

On a Mac with the Grace Notes repo, run `grace ci` (or `grace test` with the project’s usual simulator destination) to confirm the app builds and tests pass. Optionally set the device language to Turkish, toggle Today journal appearance in Settings, force-quit and relaunch, and confirm the chosen style persists.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/DesignSystem/JournalAppearanceMode.swift
+++ b/GraceNotes/GraceNotes/DesignSystem/JournalAppearanceMode.swift
@@ -8,11 +8,12 @@ enum JournalAppearanceMode: String, CaseIterable, Identifiable, Sendable {
     var id: String { rawValue }
 
     /// Interprets persisted `@AppStorage` / `UserDefaults` strings, including legacy `"summer"`.
-    /// Trims surrounding whitespace and lowercases before matching so hand-edited or oddly-cased values do not
-    /// silently fall back to Standard.
+    /// Trims surrounding whitespace and lowercases before matching so hand-edited or oddly-cased values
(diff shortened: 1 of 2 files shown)


kipyin:
Added **`GraceNotesTests/JournalAppearancePersistenceTests.swift`** so CI documents and guards journal appearance persistence: `resolveStored` is exercised for typical stored values, padded/odd casing, and the legacy **`summer` → bloom** mapping, plus **`UserDefaults`** round-trips and **`migrateLegacyJournalAppearanceRawValueIfNeeded`** using both the standard suite and isolated suites. A focused test shows **POSIX vs Turkish** lowercasing for Latin **I**, so we do not silently regress to locale-sensitive `String.lowercased()` for persisted keys. These tests run in the iOS test target on macOS CI (`grace test` / `grace ci`); they are not runnable on Linux-only agents.

PR #535, opened 2026-04-17
Align weekly reflection day counts with meaningful content

kipyin:
## Headline

Weekly review reflection days now follow the same meaningful-entry bar everywhere.

## User impact

Reflection-day totals and the week’s day activity strip stay consistent with how Grace Notes already scores a day as meaningful, so insights and completion signals are not inflated by notes-only or reflections-only stubs. Reviewers see fewer mismatches between “reflection days” and other weekly stats.

## What changed

Reflection day counting and the per-day activity set previously treated a day as active if an entry had meaningful content or only non-empty trimmed reading notes or reflections. That second path could bump reflection totals and disagree with `meaningfulEntryCount` and the rest of the weekly aggregates.

The builder now counts those days only when `hasMeaningfulContent` is true, matching the single definition used elsewhere, and removes the helper that only looked at notes and reflections text.

## Verification

On a Mac with Xcode and the project’s `grace` CLI, run `grace ci` (or `grace test` with your usual simulator destination). Optionally open Weekly Review for a week where some days have only notes or reflections and confirm reflection-day counts match expectations for meaningful entries.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Services/WeeklyReviewAggregates.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Services/WeeklyReviewAggregates.swift
@@ -140,18 +140,13 @@ private extension WeeklyReviewAggregatesBuilder {
     func reflectionDayCount(from entries: [Journal], calendar: Calendar) -> Int {
         Set(
             entries
-                .filter { $0.hasMeaningfulContent || hasReflectionSurfaceText($0) }
+                .filter(\.hasMeaningfulContent)
                 .map { calendar.startOfDay(for: $0.entryDate) }
         ).count
     }
 
     func meaningfulEntryCount(from entries: [Journal]) -> Int {
         entries.filter(\.hasMeaningfulContent).count
-    }
-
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

Thanks for the deep review—we removed the separate reflection surface-text helper so weekly reflection-day counting and rhythm “active day” logic both use `Journal.hasMeaningfulContent` alone, which matches the prior soil-row surface text plus chip progress in one place and avoids two definitions of “counted” activity. That keeps behavior the same for rows where the old helper was true while shrinking duplicate logic. We agree there’s no focused test locking `reflectionDayCount` / rhythm activity to that predicate; adding one would be a good follow-up if we want to guard against future tweaks to `hasMeaningfulContent`. The `reflectionDayCount` name still reads narrower than the predicate—that mismatch predates this change; renaming for clarity would be separate. The sparse-week path still combines reflection-day count with its own long-text thresholds as before; we didn’t change that mix in this PR.

PR #536, opened 2026-04-17
Measure journal toolbar chip titles with NSString size(withAttributes:)

kipyin:
## Headline

More reliable single-line width for journal toolbar chips by using the standard UIKit measurement path.

## User impact

Toolbar chip labels use measured widths so titles fit without awkward clipping or extra empty space. A clearer measurement path reduces the chance of subtle layout mismatches between how wide the system thinks text is and how the chip is sized.

## What changed

Single-line text width for journal toolbar chip titles no longer uses a bounding-rect pass with unlimited size and line-fragment options. It now uses the usual NSString API that reports size from the same font attributes used for drawing, which is the straightforward match for one-line labels.

Behavior stays aligned with the existing toolbar chip font setup and the extra slack for SwiftUI and CJK fallback fonts; only the core width computation was simplified.

## Verification

On a Mac with the repo and Xcode, run `swiftlint lint` from the repo root, then `grace ci` (or `grace test` with the project’s default simulator destination) to confirm the app still builds and tests pass.

## Risk

Medium

## Touch class

`ui-ux`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Views/JournalToolbarChipTitleMeasuring.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Views/JournalToolbarChipTitleMeasuring.swift
@@ -17,7 +17,14 @@ enum JournalToolbarChipTitleMeasuring {
         return metrics.scaledFont(for: semibold)
     }
 
+    /// Single-line UIKit width for toolbar chip labels using ``NSString/size(withAttributes:)``.
+    ///
+    /// Intended for **short, single-line** titles (localized completion strings). This path differs
+    /// from ``NSString/boundingRect(with:options:attributes:context:)`` (e.g. no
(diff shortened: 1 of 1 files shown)


kipyin:
Here’s a comment body you can paste:

---

**Response to review**

We switched toolbar chip title width to `NSString.size(withAttributes:)` so we measure a single line with the same font attributes we use for the chip, and we document that this helper is only for short, single-line labels. To match the review’s guidance, we added an explicit newline precondition on `singleLineTextWidth` so multiline or pasted newlines cannot silently under-measure. Call sites stay limited to localized completion titles (no intentional newlines); regression coverage still relies on existing unit tests and **macOS CI** (`grace test` / `grace ci`) because chip layout and `JournalToolbarChipTitleWidthTests` are iOS-only and do not run on Linux agents.

PR #537, opened 2026-04-17
Align Past insights refresh snapshots with calendar-day week ranges

kipyin:
## Headline

Past-tab review insights now refresh on a consistent day calendar, not mixed timestamps.

## User impact

Entries saved at different times on the same day could be treated inconsistently when deciding which journals affect Past insights and trend weeks. That could cause insights to update too often, too little, or feel out of sync with what you actually wrote. This change makes that logic line up with how the rest of the window uses calendar days.

## What changed

The code that picks which journal entries belong in the current and previous review weeks no longer uses raw date containment on the entry timestamp. It compares the entry’s calendar day (start of day) to each week range’s lower and upper bounds, the same way the past-statistics history window already works. Snapshot ordering for the refresh key now sorts entry IDs as UUIDs instead of stringifying them first, which keeps ordering stable with less overhead.

## Verification

On a Mac with the repo’s Xcode setup, run `grace ci` (or at least `grace test`) to confirm the GraceNotes scheme still builds and tests pass.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Services/ReviewInsightsRefreshPolicy.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Services/ReviewInsightsRefreshPolicy.swift
@@ -39,12 +39,13 @@ struct ReviewInsightsRefreshKey: Hashable {
         let snapshots = entries.compactMap { entry -> ReviewEntrySnapshot? in
             let entryDay = calendar.startOfDay(for: entry.entryDate)
             let inHistoryWindow = entryDay >= historyRange.lowerBound && entryDay < historyRange.upperBound
-            let inCurrentWeek = currentReviewPeriod.contains(entry.entryDate)
-            let inPreviousWeek = previousPeriod.contains(entry.entryDate)
+            let inCurrentWeek =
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

Thanks for the escalated pass—agreed there are no blocking issues.

We aligned `entrySnapshotsAffectingInsights` with the same calendar-day / half-open interval logic used for the past-statistics window, so the refresh fingerprint matches how history windows are computed instead of mixing `startOfDay` for one path with `Range.contains(entryDate)` for the weeks. Snapshot IDs are now ordered with `UUID`’s `Comparable` ordering instead of lexicographic `uuidString`, which stays fully deterministic; that will change serialized keys once on upgrade and may trigger a single extra insights refresh—data stays correct, not stale. The remaining gap you called out (week bucketing via `contains` vs normalization if `entryDate` were ever odd) is unchanged by this change; we only tightened fingerprint consistency. We did not add a test for “same local day, odd instant” since `entryDate` is expected to be normalized at write time; happy to add one if we want to lock that invariant explicitly.

PR #538, opened 2026-04-17
Resolve UI test store URL once for reset and container

kipyin:
## Headline

Keeps automated UI tests using one consistent SwiftData file when clearing and opening the store.

## User impact

Shippable app behavior for daily journaling is unchanged. This hardens UI automation and CI by preventing a rare mismatch where the reset step and the opened container could target different on-disk paths if the session marker logic ran twice under shifting conditions.

## What changed

The UI test database path is not a pure value: resolving it can read/write a marker file and choose a session key, so evaluating it twice in one setup could theoretically land on different URLs if I/O or the test environment changed between calls.

`makeForUITesting()` now resolves that URL a single time, passes it into the optional reset helper, and feeds the same URL into the SwiftData configuration. The public reset API still works as before but forwards into a private overload that deletes by an explicit URL so reset and open stay aligned.

## Verification

On macOS with Xcode, run `grace ci` from the repo root (or `grace test` targeting the UI tests that exercise `-grace-notes-reset-uitest-store`). Confirm the app builds for the simulator and UI tests pass without store-path flakiness.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Data/Persistence/SwiftData/PersistenceController.swift
+++ b/GraceNotes/GraceNotes/Data/Persistence/SwiftData/PersistenceController.swift
@@ -1,6 +1,38 @@
 import CryptoKit
 import Foundation
+import os
 import SwiftData
+
+private let persistenceUITestLogger = Logger(
+    subsystem: "com.gracenotes.GraceNotes",
+    category: "PersistenceUITest"
+)
+
+/// Deletes the UI-test SwiftData file when present. Retries once for transient file locks.
+private func removeUITestStoreFileIfPresent(at storeURL: URL) {
+    let fileManager = FileManager.default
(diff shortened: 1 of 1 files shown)


kipyin:
We now resolve the UI-test store URL once in `makeForUITesting()` and pass that same URL into both the reset path and `ModelConfiguration`, so the app never deletes one file and opens another if `uiTestStoreURL` side effects diverged between evaluations. Store removal is handled by a small helper that retries once for transient locks, then logs failures and asserts in debug builds instead of silently ignoring delete errors. The public `resetUITestStoreIfRequested()` entry still resolves the URL when called on its own, which matches today’s usage. macOS CI with `grace test` / `grace ci` remains the right place to exercise `makeForUITesting()` and store reset; marker writes still use tolerant I/O, which is an accepted residual edge case.

PR #540, opened 2026-04-17
Harden share-card redaction against invalid line indices

kipyin:
## Headline

Share images only honor redactions for lines that actually exist, and toggles cannot store invalid indices.

## User impact

Stale or out-of-range redaction indices (for example after editing or shortening journal text) could hide the wrong lines or leave privacy expectations unclear. Negative indices could also pollute stored state. The share card now treats redactions as valid only for the current line count and ignores invalid toggles.

## What changed

The redaction toggle helper now ignores negative indices so the draft never records impossible line numbers.

When building list and prose lines for the share preview, redaction is applied using only indices that fall within the current number of lines. Indices that are out of range are ignored, so the rendered card stays consistent with what is on screen and avoids mis-applying redaction across edits.

## Verification

From the repo root on macOS: run `swiftlint lint` (style). For a full simulator build aligned with CI, run `grace ci` (or `grace test` with your usual destination). Exercise the share composer: toggle redactions, edit entries so line counts change, and confirm only intended lines are hidden on the share preview.

## Risk

Medium

## Touch class

`business-logic`
---

### DIFF
--- a/GraceNotes/GraceNotes/Features/Journal/Services/ShareCardDraft.swift
+++ b/GraceNotes/GraceNotes/Features/Journal/Services/ShareCardDraft.swift
@@ -128,6 +128,7 @@ struct ShareCardDraft: Equatable, Sendable {
 
 private extension Set where Element == Int {
     mutating func toggleMembership(_ index: Int) {
+        guard index >= 0 else { return }
         if contains(index) {
             remove(index)
         } else {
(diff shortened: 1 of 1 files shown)


kipyin:
**Sentry — review feedback**

Thanks for the detailed review—we aligned share-card redaction with a single defensive rule: **only indices inside the current line count can affect what’s shown**, using `intersection(0 ..< count)` so stale or corrupted indices can’t redact a different, still-valid line (they no-op). We applied the same idea to prose so list and prose sections stay consistent, and we ignore negative indices in toggle so bad state can’t propagate. We agree **focused unit tests** for out-of-range and negative indices (e.g. `-1`, `99`, or both) would lock in those privacy expectations for future refactors; we’ll add them and rely on **`grace test` / CI on macOS** for compilation and execution. The note about prose line indices drifting if body text changes while the sheet is open matches existing behavior—we didn’t try to fix that coupling in this change.

